PR #11, opened 2019-08-21
Added a new boolean function - equals and int hashCode

4MoPDai:
There was used Class Object and construction instanceof for different exceptions. And was created hashCode

### DIFF
--- a/src/main/java/core/basesyntax/Rectangle.java
+++ b/src/main/java/core/basesyntax/Rectangle.java
@@ -36,18 +36,9 @@ public class Rectangle {
         if (obj instanceof Rectangle) {
             Rectangle rectObj = (Rectangle) obj;
 
-            if (!this.length.equals(rectObj.length)) {
-                return false;
-            }
-            if (!this.width.equals(rectObj.width)) {
-                return false;
-            }
-            if (!this.color.equals(rectObj.color)) {
-                return  false;
-            }
-            return true;
-
-
+            return this.length.equals(rectObj.length)
+                && this.width.equals(rectObj.width)
+                && this.color.equals(rectObj.color);
         }
         return false;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,29 @@ public Integer getLength() {
...
+                return false;
+            }
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * length - 13 + 2 * this.width;
```
boroda4436: @4MoPDai why you are using 31, 13 and 2 in your solution?
this is just intersting for me. Why not for example 42 24 and 5?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,29 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj instanceof Rectangle) {
```
boroda4436, line 36: @4MoPDai please use color as well as length and width

boroda4436 (changes requested):

src/main/java/core/basesyntax/Rectangle.java
4MoPDai: > @4MoPDai why you are using 31, 13 and 2 in your solution?
> this is just intersting for me. Why not for example 42 24 and 5?
        // Я использую 31, 13, 2. Потому что это уменьшает колизи.
        // то есть уменьшает возможность похожего хэш-кода у других
        //объектов

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,37 @@ public Integer getLength() {
...
+            Rectangle rectObj = (Rectangle) obj;
+
+            if (!this.length.equals(rectObj.length)) {
```
boroda4436: @4MoPDai please simplify your if statements

boroda4436 (changes requested):

PR #13, opened 2019-08-21
solution for methods equals and hashcode

TatianaPirch:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object rec) {
```
Sofasmile: `rec`? What does it mean?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+        if (this == rec) {
+            return true;
+        }
+        if (rec instanceof Rectangle) {
+            Rectangle rectangleObj = (Rectangle) rec;
+            if (!this.width.equals(rectangleObj.width)) {
```
Sofasmile: can we simplify count of `if`?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,28 @@ public Integer getLength() {
...
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj instanceof Rectangle) {
+            Rectangle rectangleObj = (Rectangle) obj;
+            if (!this.width.equals(rectangleObj.width)
```
boroda4436: simplify this this if. Just use return ...

PR #31, opened 2019-08-22
Rect

PupilDima:
Done

### DIFF
--- a/src/main/java/core/basesyntax/Rectangle.java
+++ b/src/main/java/core/basesyntax/Rectangle.java
@@ -1,6 +1,6 @@
 package core.basesyntax;
 
-import java.Objects;
+import java.util.Objects;
 
 /**
  * <p>Дан класс Rectangle. Переопределите в нем методы equals() и hashCode() так, чтобы equals()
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +18,26 @@ public Rectangle(Integer width, Integer length, String color) {
...
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj instanceof Rectangle) {
+            Rectangle rectangle = (Rectangle) obj;
+            return Objects.equals(width, rectangle.width)
+                    && Objects.equals(length, rectangle.length)
+                    && Objects.equals(color, rectangle.color);
+        } else {
```
boroda4436: remove else statement

PupilDima: Fixed.
But why it will be better?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -1,5 +1,7 @@
 package core.basesyntax;
 
+import java.util.Objects;
```
boroda4436, line 3: @PupilDima do not use Objects class

boroda4436 (changes requested):

PR #37, opened 2019-12-04
Added hw32 solution

IvanKrivoruchkoHub:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,44 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        }
+
+        if (!width.equals(recOther.getWidth())) {
+            return false;
+        }
+
+        if (!length.equals(recOther.getLength())) {
+            return false;
+        }
+        return true;
```

Tarabarnik (commented):
Do we need a lot of empty lines?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,27 @@ public Rectangle(Integer width, Integer length, String color) {
         this.color = color;
     }
 
+    public int hashCode() {
+        int result = 1;
+        result = 31 * result + (width != null ? width : 0);
```
Tarabarnik: can we use a hashcode of width and length?

Tarabarnik (approved):

PR #38, opened 2019-12-04
added overridden methods equals and hashCode

YarikJ:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,23 @@ public Integer getLength() {
...
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return Objects.equals(width, rectangle.width)
+                && Objects.equals(length, rectangle.length)
+                && Objects.equals(color, rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        return this.width * 31 + this.length + this.color.length();
```
Sofasmile: better don't use values like this: `this.width` and `this.length` and length of the color

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,25 @@ public Integer getLength() {
...
+                && Objects.equals(length, rectangle.length)
+                && Objects.equals(color, rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(width, length, color);
+    }
 }
+
```
Sofasmile, line 53: remove two blank line

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,25 @@ public Integer getLength() {
...
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return Objects.equals(width, rectangle.width)
+                && Objects.equals(length, rectangle.length)
+                && Objects.equals(color, rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(width, length, color);
```
Sofasmile: Let’s not use java.util.Objects

YarikJ:
Should I replace  calculation of hashCode with the next statement 
**return length * width * color.hashCode();** ???

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,33 @@ public Integer getLength() {
...
+        Rectangle rectangle = (Rectangle) o;
+        return this.length.equals(rectangle.length)
+                && this.width.equals(rectangle.width)
+                && (this.color == rectangle.color) || (this.color != null
+                && this.color.equals(rectangle.color));
+    }
+
+    @Override
+    public int hashCode() {
+        int colNum = 0;
```
Sofasmile: can we use a hashcode of width, length and color to calculation hashCode?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,33 @@ public Integer getLength() {
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return this.length.equals(rectangle.length)
+                && this.width.equals(rectangle.width)
+                && (this.color == rectangle.color) || (this.color != null
```
Sofasmile: why do we check only color for `null` but not checking other fields?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,28 @@ public Integer getLength() {
...
+                && this.length.equals(rectangle.length)))
+                && ((this.width == rectangle.width) || (this.width != null
+                && this.width.equals(rectangle.width)))
+                && ((this.color == rectangle.color) || (this.color != null
+                && this.color.equals(rectangle.color)));
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * length.hashCode() * width.hashCode() * color.hashCode();
```
Sofasmile: What will happen if one of field, for example length or width == 0?

Sofasmile (commented):
Good work! Make small fixing

src/main/java/core/basesyntax/Rectangle.java
YarikJ: more collisions would happen

Sofasmile (approved):

PR #39, opened 2019-12-04
hw3.2 done, please check

RomanDubovskyi:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,24 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object rectangle) {
+        if (rectangle instanceof Rectangle) {
```
Sofasmile: what is the difference between `instanceof` and `getClass()`?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,24 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object rectangle) {
```
Sofasmile, line 32: we should have four checking in this method, but now we have only two of them

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,24 @@ public Integer getLength() {
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result;
+        result = 18 * (color.length() + width + length);
```
Sofasmile: why do we use `color.length()` here?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,27 @@ public Integer getLength() {
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result;
+        result = 31 * (color.hashCode() + width + length);
```
```suggestion
        result = 31 * (color.hashCode() + width.hashCode() + length.hashCode());
```

Sofasmile (approved):

PR #40, opened 2019-12-04
HW 3.2 - solution

Julianevinska:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,23 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,23 @@ public Integer getLength() {
...
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,23 @@ public Integer getLength() {
...
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return Objects.equals(width, rectangle.width)
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,27 @@ public Integer getLength() {
...
+        }
+        Rectangle rectangle = (Rectangle) object;
+        return this.length.equals(rectangle.length)
+                && this.width.equals(rectangle.width)
+                && this.color.equals(rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * (length.hashCode() + width.hashCode()
```
Tarabarnik: length and width may be null, also. and hashcode should not be 0 or any other constant number, if some of the fields is null

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,28 @@ public Integer getLength() {
...
+        return this.length.equals(rectangle.length)
+                && this.width.equals(rectangle.width)
+                && this.color.equals(rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * ((width != null) ? width.hashCode() : 0)
+                 + ((color != null) ? color.hashCode() : 0)
+                + (Integer.toString(length).hashCode());
```
```suggestion
                + length != null ? length.hashCode() : 0);
```

Tarabarnik (approved):

PR #42, opened 2019-12-04
hw:3.2 equals and hasgCode solved

Danhov:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        Rectangle rectangle = (Rectangle) o;
+
+        if (!(this.length.equals(rectangle.length) && this.width.equals(rectangle.width)
```
Sofasmile: if (!(this.length.equals(rectangle.length) && this.width.equals(rectangle.width)
                && this.color.equals(rectangle.color))) {
            return false;
        }

        return true;
-->
return this.length.equals(rectangle.length)
&& this.width.equals(rectangle.width)
&& this.color.equals(rectangle.color);

Sofasmile (approved):

PR #43, opened 2019-12-04
solution to hw3-2

syevtushok:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,16 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public int hashCode() {
+        return 27 * getLength() * getWidth();
```
Sofasmile: how about color?
we shouldn't depend on value of fields when we find hash code

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,16 @@ public Integer getLength() {
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return this.getColor().equals(getColor())
```
Sofasmile: how about other checking? We should have 4 checking here

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public int hashCode() {
+        return 31 * getLength() * getWidth()
```
Sofasmile: what will happen if width or length ==0?
try do smth like this
https://www.baeldung.com/java-hashcode

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
...
+            * (getColor() != null ? getColor().length() : 2);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+
+        if (obj instanceof Rectangle) {
```
Sofasmile: what is the difference between method `instanceof` and `getClass()`?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
...
+            if (!this.getColor().equals(rectangle.getColor())) {
+                return false;
+            }
+            if (!this.getWidth().equals(rectangle.getWidth())) {
+                return false;
+            }
+            if (!this.getLength().equals(rectangle.getLength())) {
+                return false;
+            }
+        }
```
Sofasmile: Can we write it with && ?

src/main/java/core/basesyntax/Rectangle.java
syevtushok: Oh, I understood it. I'll change it.

src/main/java/core/basesyntax/Rectangle.java
Sofasmile: what will happen if length or width is 0?

can we use a hashcode of width, length and color and other algorithm of calculation?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,22 @@ public Integer getLength() {
...
+    public int hashCode() {
+        return 31 * getLength() * getWidth()
+            * (getColor() != null ? getColor().length() : 2);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
```
Sofasmile, line 42: how about checking on null?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,22 @@ public Integer getLength() {
...
+        Rectangle rectangle = (Rectangle) obj;
+        return ((Rectangle) obj).getClass().equals(rectangle.getClass())
+            && this.getColor().equals(rectangle.getColor())
+            && this.getWidth().equals(rectangle.getWidth())
+            && this.getLength().equals(rectangle.getLength());
+    }
+}
```
Sofasmile: code style !

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,25 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public int hashCode() {
+        return 31 * color.hashCode() * width.hashCode() * length.hashCode();
```
Sofasmile: what will happen if length or width is 0 or color = "" ?

Sofasmile (commented):
Good work!
Let's fix the latest comment :)

Sofasmile (approved):

PR #44, opened 2019-12-04
3.2 is done

dmytro-hryhoruk:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,33 @@ public Rectangle(Integer width, Integer length, String color) {
         this.color = color;
     }
 
+    @Override
+    public int hashCode() {
+        int result = 0;
+        result = color.length() + width.intValue() + length.intValue();
```
Sofasmile: is it ok to use length of fields to find hashcode?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,33 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
Sofasmile, line 27: how about checking on null?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,33 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        if (this == obj) {
+            return true;
+        }
+        if (obj instanceof Rectangle) {
+            Rectangle rectangleobj = (Rectangle) obj;
+            if (!this.color.equals(rectangleobj.color)) {
```
Sofasmile: can we simplify this three checking using && or || ?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,33 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        result = color.length() + width.intValue() + length.intValue();
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj instanceof Rectangle) {
```
Sofasmile: what is the difference between `instanceof` and `getClass()`?

src/main/java/core/basesyntax/Rectangle.java
dmytro-hryhoruk: Actually I don't know for sure but it has worked in this video https://youtu.be/Lk2vF7xz8tE?t=667

Sofasmile: which hash code we will have if we create next objects:
Rectangle rec1 = new Rectangle(20, 25, "green");
Rectangle rec2 = new Rectangle(20, 25, "black");
rec1.hachCode()   ?   rec2.hachCode() 

our hachCode() shoud be different

dmytro-hryhoruk: after Q&A session somehow my latest commit didn't get to GitHub. Now that's gonna be ok.

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,14 +16,35 @@ public Rectangle(Integer width, Integer length, String color) {
         this.color = color;
     }
 
-    public Integer getWidth() {
-        return width;
+    @Override
+    public int hashCode() {
+        int result = 0;
```
Sofasmile, line 21:
```suggestion
        int result = 31;
```
we should use some primary number)

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,14 +16,35 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        result = color.hashCode() + width.hashCode() + length.hashCode();
```
```suggestion
        result += color.hashCode() + width.hashCode() + length.hashCode();
```

Sofasmile (approved):

PR #47, opened 2019-12-04
added solution

Riptod:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -10,6 +10,29 @@
...
+        }
+        if (o instanceof Rectangle) {
+            Rectangle rectangle = (Rectangle) o;
+            if (!this.width.equals(rectangle.width)) {
+                return false;
+            }
+            if (!this.length.equals(rectangle.length)) {
+                return false;
+            }
+            return this.color.equals(rectangle.color);
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -10,6 +10,29 @@
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        return (int) (31 * width * 2 * Math.sqrt(length) + color.length());
```
Tarabarnik: Try to solve without Math library, and make it more complex to avoid collisions

Tarabarnik (commented):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -15,22 +15,22 @@ public boolean equals(Object o) {
...
+        return width.equals(rectangle.width)
+                && length.equals(rectangle.length)
+                && color.equals(rectangle.color);
     }
 
     @Override
     public int hashCode() {
-        return (int) (31 * width * 2 * Math.sqrt(length) + color.length());
+        int result = 0;
+        for (int i = 0; i < color.length(); i++) {
```
Tarabarnik: Better :) But what if width or length or color is `null`?

Tarabarnik: make a solution without loops. hint: you can use hashcodes of Rectangle's fields. And don't forget about previous comment :)

Tarabarnik (commented):
Read my comments again, please(edited)

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -10,6 +10,33 @@
...
+        Rectangle rectangle = (Rectangle) o;
+        return width.equals(rectangle.width)
+                && length.equals(rectangle.length)
+                && color.equals(rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        if (color == null || width == null || length == null) {
+            return 0;
```
Tarabarnik: we shouldn't return 0 if some field is null. The hashcode should be anyway

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -10,6 +10,28 @@
...
+        Rectangle rectangle = (Rectangle) o;
+        return width.equals(rectangle.width)
+                && length.equals(rectangle.length)
+                && color.equals(rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        final int number = 31;
+        return number * (color != null ? color.hashCode() : 0
```
Tarabarnik, line 30: what if width and/or height is null?

PR #50, opened 2019-12-04
Finished project

MrNopi:


pom.xml
```diff
@@ -25,6 +25,12 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
Sofasmile: do we need this dependency here?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,27 @@ public Rectangle(Integer width, Integer length, String color) {
         this.color = color;
     }
 
+    @Override
+    public int hashCode() {
+        return getByteSum(this.color) * this.getWidth() * this.getLength();
```
Sofasmile: try implement this method using `prime number`
and let's use hashcode of color, width and length

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +48,13 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    private int getByteSum(String str) {
```
Sofasmile: I think it's redundant :)

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,27 @@ public Rectangle(Integer width, Integer length, String color) {
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj != null && this.getClass() == obj.getClass()) {
+            Rectangle rectObj = (Rectangle)obj;
+            if (this.width - rectObj.getWidth() == 0) {
```
Sofasmile: What happen in we change `Integer width` to `String width`?
Is it correct to check difference of the values?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,27 @@ public Rectangle(Integer width, Integer length, String color) {
         this.color = color;
     }
 
+    @Override
+    public int hashCode() {
+        return 31 * color.hashCode() * this.getWidth() * this.getLength();
```
Sofasmile, line 21: what will happen if getWidth() or getLength() == 0?
and try to use hashCode() to all fields

Sofasmile: ???

Sofasmile: what about this comment?

PR #51, opened 2019-12-04
added solution for hw-3-2

AlinaPodoliaka:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj instanceof Rectangle) {
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+            }
+            return true;
+        }
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = width * length;
+        result = 31 * result + (color != null ? color.length() : 0);
```
Tarabarnik: try to make your hashcode more complex, to avoid collisions. hint: you can use hashcodes of Rectangle's fields

Tarabarnik (commented):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,28 @@ public Integer getLength() {
...
+        }
+        Rectangle rectangleObj = (Rectangle) object;
+        return rectangleObj.getWidth() == rectangleObj.width
+                && rectangleObj.getLength() == rectangleObj.length
+                && Objects.equals(rectangleObj.getColor(), rectangleObj.color);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = width.hashCode() + length.hashCode();
```
Tarabarnik: width and length also may be null

Tarabarnik (approved):

PR #52, opened 2019-12-04
equals and hashCode

graveman13:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,21 @@ public Integer getLength() {
...
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) obj;
+        return this.color.equals(rectangle.color)
+                && this.length.equals(rectangle.length)
+                && this.width.equals(rectangle.width);
+    }
+
+    public int hashCode() {
+        return ((width * 3 * 4 / 2) * (length * 4) - width * 3) / 2 + color.hashCode();
```
Tarabarnik: read about `prime number` and create a hashcode with use of this

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -43,7 +43,7 @@ public boolean equals(Object obj) {
 
     public int hashCode() {
         int resual = 31;
-        resual = resual * (width * 3 * 4 / 2) + color.hashCode() * 2;
+        resual = resual * (width * 3 * 4 / 2) + length + color.hashCode() * 2;
```
Tarabarnik: 3 * 4 / 2 - what's the magic? it is always 12, isn't it? check each field for `null` and use their hashcodes to create this one

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,24 @@ public Integer getLength() {
...
+        Rectangle rectangle = (Rectangle) obj;
+        return this.color.equals(rectangle.color)
+                && this.length.equals(rectangle.length)
+                && this.width.equals(rectangle.width);
+    }
+
+    public int hashCode() {
+        int resual = 31;
+        resual = resual + width * 3 + resual + length
+                * resual + (color == null ? 3 : color.hashCode());
```
Tarabarnik: good. what about other fields?

Tarabarnik (approved):

PR #53, opened 2019-12-04
added solution method

tsehelnyk:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,33 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj instanceof Rectangle) {
```
Tarabarnik: try to use getClass() not instanceof

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,33 @@ public Integer getLength() {
...
+            if (!this.getWidth().equals(rectangle.getWidth())) {
+                return false;
+            }
+            if (!this.getLength().equals(rectangle.getLength())) {
+                return false;
+            }
+            if (!this.getColor().equals(rectangle.getColor())) {
+                return false;
+            }
+        }
```

Tarabarnik (commented):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+        if (obj == null) {
+            return false;
+        }
+        if (this.getClass() == obj.getClass()) {
+            Rectangle rectangle = (Rectangle) obj;
+            return ((this.getWidth().equals(rectangle.getWidth()))
+                    && (this.getLength().equals(rectangle.getLength()))
+                    && (this.getLength().equals(rectangle.getLength()))
+                    && (this.getColor().equals(rectangle.getColor()))) ? true : false;
+        } else {
```
Tarabarnik: `else` is redundant here

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+            return false;
+        }
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = 31;
+        hash = hash * Integer.parseInt(Integer.toBinaryString(this.length + 31));
+        hash = hash * Integer.parseInt(Integer.toBinaryString(this.width + 31));
+        hash = hash * Integer.parseInt(Integer.toBinaryString(this.color.charAt(1) + 31));
```
Tarabarnik: what will happen if any of the fields is null?

Tarabarnik: hint: try to use hashcode of the fields to create this one

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
...
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (this.getClass() == obj.getClass()) {
+            Rectangle rectangle = (Rectangle) obj;
+            return ((this.getWidth().equals(rectangle.getWidth()))
```
Tarabarnik: no need to use getters while you are in the class which has this field

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
...
+                    && (this.getLength().equals(rectangle.getLength()))
+                    && (this.getLength().equals(rectangle.getLength()))
```
Tarabarnik: Why checking length twice?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
...
+                    && (this.getColor().equals(rectangle.getColor()))) ? true : false;
```
Tarabarnik: remove ternary operator. it's redundant

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = 31;
+        hash = hash * this.length.hashCode();
+        hash = hash * this.width.hashCode();
+        hash = hash * this.color.hashCode();
```
Tarabarnik: what if any field is null?

Tarabarnik (approved):

PR #55, opened 2019-12-04
Well, overriding was done automatically by IJI

IegorSoboliev:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -1,5 +1,7 @@
 package core.basesyntax;
 
+import java.util.Objects;
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,25 @@ public Integer getLength() {
...
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return width.equals(rectangle.width)
+                &&
```
Tarabarnik: there is no need to keep this operators at the separate line

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,23 @@ public Integer getLength() {
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return width.equals(rectangle.width) &&
```
```suggestion
        return width.equals(rectangle.width)
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,23 @@ public Integer getLength() {
...
+                length.equals(rectangle.length) &&
```
```suggestion
                && length.equals(rectangle.length) 
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,23 @@ public Integer getLength() {
...
+                color.equals(rectangle.color);
```
```suggestion
                && color.equals(rectangle.color);
```

Tarabarnik (commented):
try that way :)

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,23 @@ public Integer getLength() {
...
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return width.equals(rectangle.width)
+                && length.equals(rectangle.length)
+                && color.equals(rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * (width + length) + color.length();
```
Tarabarnik: this hashCode is very simple and may cause lots of collisions. Try to complicate it with hashcodes of Rectangle's fields. and don't forget, that they might be null :)

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,24 @@ public Integer getLength() {
...
+        Rectangle rectangle = (Rectangle) o;
+        return width.equals(rectangle.width)
+                && length.equals(rectangle.length)
+                && color.equals(rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * ((width != null) ? width.hashCode() : 0)
+                + 31 * ((length != null) ? length.hashCode() : 0) + color.length();
```
Tarabarnik: color is also a field of Rectangle class. so if you ask its method without null check, you may get NullPointerException

IegorSoboliev: I had thought about it:(. Thank you!

PR #56, opened 2019-12-04
added a solution for equals and hashcode homework

Prox1k:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -1,5 +1,7 @@
 package core.basesyntax;
 
+import java.util.Objects;
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,32 @@ public Integer getLength() {
...
+        if (!this.color.equals(rectangle.color)) {
+            return false;
+        }
+        if (!this.width.equals(rectangle.width)) {
+            return false;
+        }
+        if (!this.length.equals(rectangle.length)) {
+            return false;
+        }
+        return true;
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,28 @@ public Integer getLength() {
...
+        Rectangle rectangle = (Rectangle) obj;
+        return width.equals(rectangle.width) && length.equals(rectangle.length)
+                && Objects.equals(color, rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 17;
+        result = 37 * result + (color == null ? 0 : color.hashCode());
+        result = 37 * result + width;
```
Tarabarnik: what if width or length is null?

Tarabarnik:
java.util.Objects. don't use it, please

Tarabarnik (approved):

PR #60, opened 2019-12-05
Added solution for Mate HW3.2

AlexanderSashchenko:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,24 @@ public Integer getLength() {
...
+        if (this == object) {
+            return true;
+        }
+        if (object == null || object.getClass() != this.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
+        return this.getWidth().equals(rectangle.getWidth())
+                && this.getLength().equals(rectangle.getLength())
+                && this.getColor().equals(rectangle.getColor());
```
Tarabarnik: there is no need to use getters while you are in a class which has this field

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,24 @@ public Integer getLength() {
...
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * this.getWidth() + 31 * this.getLength() + 31 * this.getColor().hashCode();
```
Tarabarnik: what if any field is null?

src/main/java/core/basesyntax/Rectangle.java
AlexanderSashchenko: Ok, fixed.

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -37,14 +37,14 @@ public boolean equals(Object object) {
...
-                && this.getColor().equals(rectangle.getColor());
+        return this.width.equals(rectangle.width)
+                && this.length.equals(rectangle.length)
+                && this.color.equals(rectangle.color);
     }
 
     @Override
     public int hashCode() {
-        return 31 * this.getWidth() + 31 * this.getLength() + 31 * this.getColor().hashCode();
+        return 31 * this.width + 31 * this.length + 31 * this.color.hashCode();
```
Tarabarnik: what will happen if any of your fields is null?

src/main/java/core/basesyntax/Rectangle.java
AlexanderSashchenko: Also fixed.

Tarabarnik (approved):

PR #63, opened 2019-12-05
hm3.2 v1

Levkosvd:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +28,28 @@ public Integer getLength() {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+
+        if (obj instanceof Rectangle) {
```
Tarabarnik: we should use getClass() here

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +28,28 @@ public Integer getLength() {
...
+            Rectangle rectangle = (Rectangle) obj;
+            return this.length.equals(rectangle.length) && this.width.equals(rectangle.width)
+                    && this.color.equals(rectangle.color);
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * ((width != null) ? width.hashCode() : 0)
```
Tarabarnik: what will happen if u multiply a zero?

PR #66, opened 2020-04-03
homework 5_2

ZubenkoHalyna:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,27 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || obj.getClass() != Rectangle.class) {
```
```suggestion
        if (obj == null || obj.getClass() != getClass()) {
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,27 @@ public Integer getLength() {
...
+        return width.equals(rectangle.width)
+                && length.equals(rectangle.length)
+                && color.equals(rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 17;
+        result = 31 * result + width;
+        result = 31 * result + length;
```
Tarabarnik: can we use hashCodes of width and length?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,27 @@ public Integer getLength() {
...
+        result = 31 * result + color.hashCode();
```
Tarabarnik: what's gonna happen if color == null?

src/main/java/core/basesyntax/Rectangle.java
ZubenkoHalyna: Technically yes, because it's a wrapper class Integer and it has hashCode method. But we shouldn't do it because Integer.hashCode will return unique value for two unique Integer objects even if they has the same values. As we use equals for this fields in Rectangle.equals method we shouldn't use hashCode method for this fields here. It leads to the situation when two equal objects will have different hashCodes.

src/main/java/core/basesyntax/Rectangle.java
ZubenkoHalyna: NPE, but length and width also may cause NPE both in equals and hashCode method. Should I add additional checks?

Tarabarnik (approved):

PR #70, opened 2020-04-03
Added solution

Allduiin:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,26 @@ public Integer getLength() {
...
+            return width.equals(rectangle.width) && length.equals(rectangle.length)
+                    && color.equals(rectangle.color);
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 17;
+        result += 31 * result + width.hashCode();
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,26 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj.getClass().equals(Rectangle.class)) {
```
dmytro-hryhoruk, line 39: what if obj would be null?

dmytro-hryhoruk (approved):

PR #72, opened 2020-04-03
added solution for homework 6.2

vladpetrovskyi:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (obj.getClass() == Rectangle.class) {
```
dmytro-hryhoruk: if (obj == null || getClass() != obj.getClass()) {
            return false;
 }

dmytro-hryhoruk (approved):

PR #73, opened 2020-04-03
equals and hashcode

Vinglish:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +18,22 @@ public Rectangle(Integer width, Integer length, String color) {
         this.color = color;
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this != o && getClass() != o.getClass()) {
```
dmytro-hryhoruk: what about o == null?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +18,22 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return Objects.equals(width, rectangle.width)
+                && Objects.equals(length, rectangle.length)
+                && Objects.equals(color, rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(width, length, color);
```
dmytro-hryhoruk: write your own implementation using primary number 31

src/main/java/core/basesyntax/Rectangle.java
Vinglish: I tried to do this but InteliJ told me "this != o" covered "o == null".  to add anyway?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +18,22 @@ public Rectangle(Integer width, Integer length, String color) {
         this.color = color;
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this != o && getClass() != o.getClass()) {
+            return false;
+        }
```
dmytro-hryhoruk, line 28: let's check if (this == o) return true;

src/main/java/core/basesyntax/Rectangle.java
Vinglish: Done

src/main/java/core/basesyntax/Rectangle.java
Vinglish: Done

Sofasmile (approved):

PR #75, opened 2020-04-03
added solution of hw-6.2

KseniiaMakarova:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,27 @@ public Integer getLength() {
...
+        if (this == object) {
+            return true;
+        }
+        if (object == null || getClass() != object.getClass()) {
+            return false;
+        }
+        Rectangle other = (Rectangle) object;
+        return getWidth().equals(other.getWidth())
+                && getLength().equals(other.getLength())
+                && getColor().equals(other.getColor());
```
dmytro-hryhoruk: do we need to use getters here?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,27 @@ public Integer getLength() {
...
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 13;
+        result = result * 31 + getWidth().hashCode();
+        result = result * 31 + getLength().hashCode();
+        result = result * 31 + getColor().hashCode();
+        return result;
```

KseniiaMakarova: NPE :(
fixed this)

src/main/java/core/basesyntax/Rectangle.java
KseniiaMakarova: no - we are in the same class
fixed)

dmytro-hryhoruk (approved):

PR #77, opened 2020-04-03
hw6.2 solution

Andrewmazyar:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,32 @@ public Integer getLength() {
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 11;
+        result = 12 * width.hashCode();
+        result = 12 * length.hashCode();
+        result = 12 * color.hashCode();
```
dmytro-hryhoruk: what if some of the fields would be null?

Andrewmazyar: will be panic))))

dmytro-hryhoruk (approved):

PR #82, opened 2020-04-03
Added solution.

Merkald:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,32 @@ public Rectangle(Integer width, Integer length, String color) {
...
 
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (obj.getClass().equals(Rectangle.class)) {
```
```suggestion
        if (obj.getClass().equals(getClass())) {
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,32 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 13;
+        result = 31 * result + width.hashCode();
+        result = 31 * result + length.hashCode();
+        result = 31 * result + color.hashCode();
```
Tarabarnik: що буде, якщо якесь із полів == null?

Merkald: NullPointerException зараз переправлю

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,38 @@ public Rectangle(Integer width, Integer length, String color) {
...
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (obj.getClass().equals(Rectangle.class)) {
+            Rectangle rectangle = (Rectangle) obj;
+            return width.equals(((Rectangle) obj).width)
```
RomanDubovskyi: do we need to cast every field of obj, if we have already rectangle object, can we use it here?

src/main/java/core/basesyntax/Rectangle.java
RomanDubovskyi: have you fixed it ?

RomanDubovskyi (approved):

PR #84, opened 2020-04-03
added solution

6ALLIKA:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,32 @@ public Rectangle(Integer width, Integer length, String color) {
...
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null) {
+            return false;
+        }
+        if (o.getClass().equals(Rectangle.class)) {
```
```suggestion
        if (o.getClass().equals(getClass())) {
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,32 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 7;
+        result = 17 * result + color.hashCode();
+        result = 17 * result + width;
+        result = 17 * result + length;
+        return result;
```

6ALLIKA: good call, it would be NPE

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,32 @@ public Rectangle(Integer width, Integer length, String color) {
...
+                    && color.equals(rectangle.color);
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 7;
+        result = (color == null) ? 17 * result : 17 * result + color.hashCode();
+        result = (width == null) ? 17 * result : 17 * result + width;
```
```suggestion
        result = (width == null) ? 17 * result : 17 * result + width.hashCode();
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,32 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        result = (length == null) ? 17 * result : 17 * result + length;
```
```suggestion
        result = (length == null) ? 17 * result : 17 * result + length.hashCode();
```

Sofasmile (approved):

PR #87, opened 2020-04-03
Hw 6 2

IrinaDyakonova:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) {
+            return false;
```
RomanDubovskyi: should we return false if we are comparing object with itself?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+        }
+        if (object == null) {
+            return false;
+        }
+        if (object.getClass().equals(Rectangle.class)) {
```
RomanDubovskyi:
```suggestion
        if (object.getClass().equals(getClass())) {
```

src/main/java/core/basesyntax/Rectangle.java
IrinaDyakonova: no(

RomanDubovskyi (approved):

PR #89, opened 2020-04-03
Solution hw62

woododoCode:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,27 @@ public Integer getLength() {
...
+
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+        if (object == null) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
```
Sofasmile: how about checking on `getClass`?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,27 @@ public Integer getLength() {
...
+        return rectangle.width.equals(width)
+                && rectangle.length.equals(length)
+                && rectangle.color.equals(color);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 11;
+        result = 23 * result + width.hashCode();
```
Sofasmile: what will happen if fields are null?

Sofasmile (approved):

PR #93, opened 2020-04-03
Added solution to the 2nd task

AnastasiiaKosiak:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+        if (object == null) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle)object;
+        if (object.getClass().equals(Rectangle.class)) {
+            return width.equals(rectangle.width)
+                    && length.equals(rectangle.length)
+                    && color.equals(rectangle.color);
+        }
+        if (this == object) {
```
RomanDubovskyi: use this check at the beginning of the method, it will be better for performance

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 17;
+        result = 31 * result + width.hashCode();
```

RomanDubovskyi (approved):

PR #97, opened 2020-04-03
override equals & hashCode methods

olehedza:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,27 @@ public Integer getLength() {
...
+        Rectangle rectangle = (Rectangle) r;
+        return width.equals(rectangle.width)
+                && length.equals(rectangle.length)
+                && color.equals(rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 13;
+        result *= 31 + width.hashCode();
```

olehedza: Checking for null was added

RomanDubovskyi (approved):

PR #100, opened 2020-04-04
added solution

Tonio-an:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+                    && length.equals(rectangle.length)
+                    && color.equals(rectangle.color);
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 17;
+        result = 31 * width.hashCode();
```
RomanDubovskyi, line 53: what happens, if one of the fields == null ?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+
+        if (object == null) {
+            return false;
+        }
+
+        if (object.getClass().equals(Rectangle.class)) {
```
RomanDubovskyi, line 41:
```suggestion
        if (object.getClass().equals(getClass())) {
```

PR #102, opened 2020-04-04
Done

vovkandrew:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int generatedHashCode = 9;
+        generatedHashCode = generatedHashCode * 19 + width;
+        generatedHashCode = generatedHashCode * 19 + length;
+        generatedHashCode = generatedHashCode * 19 + color.hashCode();
```
RomanDubovskyi: what happens, if color == null ?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+        if (object == null) {
+            return false;
+        }
+        if (object.getClass().equals(Rectangle.class)) {
```
RomanDubovskyi:
```suggestion
        if (object.getClass().equals(getClass())) {
```

src/main/java/core/basesyntax/Rectangle.java
vovkandrew: then we will add just 0 as hashCode(null) = 0, is that prohibited?

RomanDubovskyi: you sure that there won't be nullPointerException if color == null? 
it's not hashCode(null) it's null.hashCode()

vovkandrew: null check added, I think that may help

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,33 @@ public Integer getLength() {
...
+
+    @Override
+    public int hashCode() {
+        int generatedHashCode = 9;
+        generatedHashCode = generatedHashCode * 19 + width;
+        generatedHashCode = generatedHashCode * 19 + length;
+        if (color == null) {
+            generatedHashCode = generatedHashCode * 19;
+        }
+        generatedHashCode = generatedHashCode * 19 + color.hashCode();
```
RomanDubovskyi: did you test your current solution with color = null? 
cause line after if statement is being executed anyway

vovkandrew: Oh, missed that, have already fixed!

RomanDubovskyi (approved):

PR #103, opened 2020-04-04
added equals() and hashcode()

akutil:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return width != null && length != null && color != null
```
RomanDubovskyi, line 39: what will be result if you comparing two same objects, that both have their color == null and other fields are OK ?

akutil: I understood. If we use my code we will receive false.

RomanDubovskyi (approved):

PR #106, opened 2020-04-05
Methods hashCode and equals are overridden

NikolayAbysov:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,21 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public int hashCode() {
+        int result = 42;
+        result = result * 32 * width.hashCode();
+        result = result * 32 * length.hashCode();
+        result = result * 32 * color.hashCode();
```
RomanDubovskyi: check for null before calling "field.hashcode()", cause you can get nullPointer in some cases

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,21 @@ public Integer getLength() {
...
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj.getClass() == this.getClass()
```
RomanDubovskyi: there should be 4 checks, you have currently only 2

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public int hashCode() {
+        if (width != null && length != null && color != null) {
```
RomanDubovskyi: each check should be separate, cause you can have situation, where only one field is null and other are ok, so you want to take them into account

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (obj.getClass().equals(this.getClass())) {
+            return ((Rectangle) obj).getColor().equals(this.color)
```
RomanDubovskyi: nice, but better to cast whole object to Rectangle once, than you won't need to cast each field separately

RomanDubovskyi (approved):

PR #108, opened 2020-04-05
added 2 overrided methods to Rectangle class

MarberyUA:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,20 @@ public Integer getLength() {
...
+    public boolean equals(Object obj) {
+        return obj instanceof Rectangle && (this.getColor().equals(((Rectangle) obj).getColor())
+                && this.getLength().equals(((Rectangle) obj).getLength())
+                && this.getWidth().equals(((Rectangle) obj).getWidth()));
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 3;
+        result = result * 11 + width.hashCode();
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,20 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof Rectangle && (this.getColor().equals(((Rectangle) obj).getColor())
```
RomanDubovskyi: should we use it instanceof or getClass() and why? 
+ there should be 4 checks in total, you have 2 now

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,39 @@ public Integer getLength() {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+
+        if (obj == null) {
+            return false;
+        }
+        if (!(obj instanceof Rectangle)) {
```
RomanDubovskyi: should we use here getClass() or instanceof ? 
and why

MarberyUA: We should use getClass() because we can get troubles when this class get child class

RomanDubovskyi (approved):

PR #110, opened 2020-04-05
added solution

Sumned:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,26 @@ public Integer getLength() {
...
+        }
+        return width.equals(((Rectangle) obj).width)
+                && length.equals(((Rectangle) obj).length)
+                && color.equals(((Rectangle) obj).color);
+    }
+
+    @Override
+    public int hashCode() {
+        int primary = 42;
+        int result = primary + color.hashCode();
```
RomanDubovskyi: don't we need check this field for null as well?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,26 @@ public Integer getLength() {
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null || obj.getClass() != this.getClass()) {
+            return false;
+        }
+        if (obj == this) {
+            return true;
+        }
+        return width.equals(((Rectangle) obj).width)
```
RomanDubovskyi: better to cast whole object first and then you can compare its fields without casting them

RomanDubovskyi (approved):

PR #115, opened 2020-04-05
Added solution for Task 6.2

Filipp888:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,33 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        }
+
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 17;
+        result = 31 * result + width.hashCode();
+        result = 31 * result + length.hashCode();
```
RomanDubovskyi: check for null before calling "field.hashcode()", cause you can get nullPointer in some cases, it's better to avoid it

RomanDubovskyi (approved):

PR #122, opened 2020-08-03
solution added

margarita-ignatieva:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -17,6 +12,7 @@ public Rectangle(Integer width, Integer length, String color) {
     }
 
     public Integer getWidth() {
+
```
```suggestion
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +23,41 @@ public Integer getLength() {
...
+            return false;
+        }
+
+        if (obj.getClass() != this.getClass()) {
+            return false;
+        }
+
+        Rectangle rectangle = (Rectangle) obj;
+        if ((width == rectangle.width) && (length != rectangle.length)
+                && (color.equals(rectangle.getColor()))) {
```
dmytro-hryhoruk: so it can return true even if length doesn't equal?

dmytro-hryhoruk: what if color would be null?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +23,41 @@ public Integer getLength() {
...
+            return true;
+        }
+
+        if ((width == null && rectangle.width == null)
+                && (length == null && rectangle.length == null)
+                && (color == null && rectangle.color == null)) {
+            return true;
```
dmytro-hryhoruk: can we simplify that?

src/main/java/core/basesyntax/Rectangle.java
margarita-ignatieva: done

src/main/java/core/basesyntax/Rectangle.java
margarita-ignatieva: done

src/main/java/core/basesyntax/Rectangle.java
margarita-ignatieva: done

dmytro-hryhoruk (approved):

PR #124, opened 2020-08-03
added solution

tarkaiev:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,33 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) o;
+        if (rectangle.color == null || rectangle.width == null || rectangle.length == null
+                || color == null || width == null || length == null) {
+            return this.hashCode() == rectangle.hashCode();
+        }
+        return width.equals(rectangle.width) && length.equals(rectangle.length)
+                && color.equals(rectangle.color);
```
KseniiaMakarova: see checklist: While comparing fields at equals, use single return statement for all fields(e.g. return check1 && check2 && check3;).

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,33 @@ public Rectangle(Integer width, Integer length, String color) {
...
+    }
+
+    @Override
+    public int hashCode() {
+        int primecode = 31;
+        int result = 1;
+        result = primecode * result + (width == null ? 0 : width);
+        result = primecode * result + (length == null ? 0 : length);
```
KseniiaMakarova: see checklist: you need to use hashcodes of fields

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
tarkaiev: thought that need hashcodes only in non-numeric fields
fixed

src/main/java/core/basesyntax/Rectangle.java
tarkaiev: fixed

KseniiaMakarova (approved):

PR #125, opened 2020-08-03
Added solution 6.2

DmGrt:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +22,27 @@ public Integer getLength() {
...
+            return true;
+        }
+        if (object == null || getClass() != object.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
+        return width == null && rectangle.width == null
+                || width != null && rectangle.width.equals(width)
+                && length == null && rectangle.length == null
+                || color != null && rectangle.color.equals(color);
```
dmytro-hryhoruk: try to test it when some of yours object field will be null

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +22,27 @@ public Integer getLength() {
...
+    }
+
+    @Override
+    public int hashCode() {
+        int primaryNumber = 14;
```
dmytro-hryhoruk: is 14 a primary number?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +22,29 @@ public Integer getLength() {
...
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+        if (object == null || getClass() != object.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
+        return (length == null || rectangle.length == null)
```
```suggestion
        return length == null
```this condition should be enough

dmytro-hryhoruk (approved):

PR #126, opened 2020-08-03
Added solution to HW6-2

MichaelShevchenko:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,26 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object other) {
+        if (other != null && getClass() == other.getClass() && this == other) {
```
KseniiaMakarova: these should be separate checks
they make no sense if put all together

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,26 @@ public Integer getLength() {
...
+
+        Rectangle rect = (Rectangle) other;
+        return (width == rect.width || (width != null && width.equals(rect.getWidth())))
+                && (length == rect.length || (length != null && length.equals(rect.getLength())))
+                && (color == rect.color || (color != null && color.equals(rect.getColor())));
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 42;
```
KseniiaMakarova: is it a prime number?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
MichaelShevchenko: done

src/main/java/core/basesyntax/Rectangle.java
MichaelShevchenko: done

KseniiaMakarova (approved):

PR #127, opened 2020-08-03
My solution

frostpv:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) o;
+        if (width != null ? !width.equals(rectangle.width) : rectangle.width != null) {
+            return false;
+        }
+        if (length != null ? !length.equals(rectangle.length) : rectangle.length != null) {
+            return false;
+        }
+        return color != null ? color.equals(rectangle.color) : rectangle.color == null;
```
KseniiaMakarova: nice!

let's see our checklist and try refactoring this: `While comparing fields at equals, use single return statement for all fields(e.g. return check1 && check2 && check3;).`

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,26 @@ public Integer getLength() {
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return (width != null ? !width.equals(rectangle.width) : rectangle.width != null)
+                || (length != null ? !length.equals(rectangle.length) : rectangle.length != null)
+                || (color != null ? color.equals(rectangle.color) : rectangle.color == null);
```
KseniiaMakarova: let's refactor this
for each field, we should return `true` 
if `this.field` and `rectangle.field` are both null 
OR if `this.field` is not null AND is equal to `rectangle.field`

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #130, opened 2020-08-03
ready solution hw-6.2

RomanHreits:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,38 @@ public Integer getLength() {
...
+
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+        if (object == null) {
+            return false;
+        }
+        if (object instanceof Rectangle) {
```
dmytro-hryhoruk: why have you choose to use instanceof instead of getClass()?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,38 @@ public Integer getLength() {
...
+                    && color.equals(rectangle.getColor())))) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 9;
```
dmytro-hryhoruk: why 9?

RomanHreits: only prime numbers...it is a mistake
done

src/main/java/core/basesyntax/Rectangle.java
RomanHreits: done

dmytro-hryhoruk (approved):

PR #133, opened 2020-08-03
HOMEWORK-6-2: - done;

jah701:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,34 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        int res = 17;
+        res = 31 * res + width.hashCode();
+        res = 31 * res + length.hashCode();
+        res = 31 * res + color.hashCode();
+        return res;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (width == null || length == null || color == null || obj == null) {
```
dmytro-hryhoruk: why it should be true?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,34 @@ public Rectangle(Integer width, Integer length, String color) {
...
+            return true;
+        } else if (obj == this) {
+            return true;
+        } else if (obj.getClass().equals(Rectangle.class)) {
```
dmytro-hryhoruk: what if object == null?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,34 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        res = 31 * res + length.hashCode();
+        res = 31 * res + color.hashCode();
+        return res;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (width == null || length == null || color == null || obj == null) {
+            return true;
+        } else if (obj == this) {
```
dmytro-hryhoruk: we don't have to use else block after return statement cause it will exit from method

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,38 @@ public Rectangle(Integer width, Integer length, String color) {
...
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (obj.getClass().equals(Rectangle.class)) {
+            Rectangle rectangle = (Rectangle) obj;
+            return width == null && rectangle.width == null
```
dmytro-hryhoruk: what if only width == null, but rectangle.width don't. What will happen when it will check the equals part?

dmytro-hryhoruk (approved):

PR #134, opened 2020-08-03
hw6-2 is done

vladkuzya:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,42 @@ public Rectangle(Integer width, Integer length, String color) {
...
+            return false;
+        }
+
+        Rectangle rectangle = (Rectangle) obj;
+        if (obj.getClass().equals(Rectangle.class)
+                && (width != null || length != null || color != null)) {
+            return width.equals(rectangle.width)
+                    && length.equals(rectangle.length)
+                    && color.equals(rectangle.color);
+        } else {
```
dmytro-hryhoruk: return statement will exit from this method, so there is no need to use else after that

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,42 @@ public Rectangle(Integer width, Integer length, String color) {
...
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+
+        if (obj == null) {
+            return false;
+        }
+
+        Rectangle rectangle = (Rectangle) obj;
```
dmytro-hryhoruk, line 38: You are casting object to some class, but at this point you don't know if it has the same type or not. What exception you can get in this case?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,42 @@ public Rectangle(Integer width, Integer length, String color) {
...
+        if (obj.getClass().equals(Rectangle.class)
+                && (width != null || length != null || color != null)) {
+            return width.equals(rectangle.width)
+                    && length.equals(rectangle.length)
+                    && color.equals(rectangle.color);
+        } else {
+            return width == rectangle.width
+                    && length == rectangle.length
+                    && color == rectangle.color;
```
dmytro-hryhoruk: so if getClass() gave different results for these objects, we still have a chance to get true from equals()?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,42 @@ public Rectangle(Integer width, Integer length, String color) {
...
+            return true;
+        }
+
+        if (obj == null) {
+            return false;
+        }
+
+        Rectangle rectangle = (Rectangle) obj;
+        if (obj.getClass().equals(Rectangle.class)
+                && (width != null || length != null || color != null)) {
```
dmytro-hryhoruk: If some object has null field values it doesn't mean that it cannot be equal to other object

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,39 @@ public Rectangle(Integer width, Integer length, String color) {
...
+            return true;
+        }
+
+        if (obj == null || !obj.getClass().equals(Rectangle.class)) {
+            return false;
+        }
+
+        try {
+            Rectangle rectangle = (Rectangle) obj;
+            return (width == null || rectangle.width == null) ? width == rectangle.width
```
```suggestion
            return width == null  ? width == rectangle.width
```
that should be enough.

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,6 +16,39 @@ public Rectangle(Integer width, Integer length, String color) {
...
+                    : width.equals(rectangle.width)
+                    && (length == null || rectangle.length == null) ? length == rectangle.length
+                    : length.equals(rectangle.length)
+                    && (color == null || rectangle.color == null) ? color == rectangle.color
+                    : color.equals(rectangle.color);
+        } catch (ClassCastException e) {
+            e.printStackTrace();
+        }
```
dmytro-hryhoruk: 1)We don't catch RuntimeExceptions(Which ClassCastException is). Mostly the occur due to a programmer mistake in code. 
2) Don't ignore exceptions if you are catching them(rethrow them);
3) !obj.getClass().equals(Rectangle.class) 
this check guarantees that such exception will never occur&

dmytro-hryhoruk (approved):

PR #136, opened 2020-08-03
added solution

Victoria-Moiseienko:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,46 @@ public Integer getLength() {
...
+
+        if (this == object) {
+            return true;
+        }
+
+        if (null == object) {
+            return false;
+        }
+
+        if (object instanceof Rectangle) {
```
dmytro-hryhoruk: what is the difference between instanceof and getClass()?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,46 @@ public Integer getLength() {
...
+            Rectangle rectangle = (Rectangle) object;
+            return checkNullOrEquals(width, rectangle.getWidth())
```
dmytro-hryhoruk: let's not create other private methods for equals(). In this case it's better to have it in your method

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,46 @@ public Integer getLength() {
...
+                    &&
+                    checkNullOrEquals(length, rectangle.getLength())
+                    &&
+                    color == rectangle.getColor();
+        }
+        return false;
+    }
+
+    private boolean checkNullOrEquals(Integer a, Integer b) {
+        return (null == a && null == b || a.equals(b));
```
dmytro-hryhoruk: what if only a == null, but b don't. What will happen when it will check the right part?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,46 @@ public Integer getLength() {
...
+        int result = 13;
+        if (null != width) {
+            result += 7 * width;
+        }
+        if (null != length) {
+            result += 3 * length;
+        }
+        if (null != color) {
+            result += 11 * color.hashCode();
+        }
```
dmytro-hryhoruk: choose one prime number and use it for each field

src/main/java/core/basesyntax/Rectangle.java
Victoria-Moiseienko: Fixed. Done

src/main/java/core/basesyntax/Rectangle.java
Victoria-Moiseienko: Done

src/main/java/core/basesyntax/Rectangle.java
Victoria-Moiseienko: Updated. Done

src/main/java/core/basesyntax/Rectangle.java
Victoria-Moiseienko: Done

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,42 @@ public Integer getLength() {
...
+            return true;
+        }
+
+        if (null == object) {
+            return false;
+        }
+
+        if (object.getClass().equals(Rectangle.class)) {
+            Rectangle rectangle = (Rectangle) object;
+            return ((null == width && null == rectangle.getWidth())
```
dmytro-hryhoruk: still the same mistake as was in private method

Victoria-Moiseienko: Fixed

```suggestion
            return ((width == rectangle.getWidth())
```
this should be enough

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,42 @@ public Integer getLength() {
...
+                            || (null != width && width.equals(rectangle.getWidth())))
```
```suggestion
                            || (width != null && width.equals(rectangle.getWidth())))
```
easier to read)

dmytro-hryhoruk (approved):

PR #138, opened 2020-08-03
Added solution

AndrewPolishchuc:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +28,34 @@ public Integer getLength() {
...
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (obj.getClass().equals(Rectangle.class)) {
+            Rectangle rectangle = (Rectangle) obj;
+            return (color != null && color.equals(rectangle.color)
+                    || (rectangle.color == null && color == null))
```
```suggestion
                    || rectangle.color == color)
```
just a note: this can be simplified

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +28,34 @@ public Integer getLength() {
...
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 0;
+        result = width != null ? HASHCODENUMBER * result + width.hashCode() : HASHCODENUMBER;
+        result = length != null ? HASHCODENUMBER * result + length.hashCode()
+                : HASHCODENUMBER * result;
+        result = color != null ? HASHCODENUMBER * result + color.hashCode() : HASHCODENUMBER % 10;
```
KseniiaMakarova: what happens when we multiply by 0 on line 55?
also, let's use the same calculation logic for all the fields

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
AndrewPolishchuc: Sure. Done.

src/main/java/core/basesyntax/Rectangle.java
AndrewPolishchuc: Done)

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -6,6 +6,7 @@
  * а hashCode() был бы согласованным с реализацией equals().</p>
  */
 public class Rectangle {
+    private static final int HASHCODENUMBER = 52;
```
KseniiaMakarova: let's use prime numbers

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +28,36 @@ public Integer getLength() {
...
+                    || (rectangle.length == length));
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 0;
+        result = width != null ? HASHCODENUMBER * result + width.hashCode()
+                : HASHCODENUMBER * result;
```
KseniiaMakarova: HASHCODENUMBER * result  will give us 0
is it supposed to be like that? maybe it's better to change result's initial value?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
AndrewPolishchuc: You are right, a number divisible by 2 is not the best choice. The best choice is 31)

src/main/java/core/basesyntax/Rectangle.java
AndrewPolishchuc: There should have been "result +=", but I wrote a better version. Done)

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +28,33 @@ public Integer getLength() {
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 0;
+        result *= HASHCODENUMBER + (width != null ? width.hashCode() : 0);
+        result *= HASHCODENUMBER + (color != null ? color.hashCode() : 0);
+        result *= HASHCODENUMBER + (length != null ? length.hashCode() : 0);
```
KseniiaMakarova, line 57: we are multiplying `result` on each line
`result` value is 0
what do we get if we multiply any value by 0?

please reply to this comment below

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
AndrewPolishchuc: We'll get eternal zero. I did not initialize correctly result.  I had to initialize "result" with any positive prime number.

KseniiaMakarova (approved):

PR #139, opened 2020-08-03
added solution 6.2

TimoxaV:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,33 @@ public Integer getLength() {
...
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || !obj.getClass().equals(Rectangle.class)) {
+            return false;
+        }
+        Rectangle objectCast = (Rectangle) obj;
+        if (length == null && width == null && color == null && objectCast.color == null
+                && objectCast.width == null && objectCast.length == null) {
+            return hashCode() == objectCast.hashCode();
```
dmytro-hryhoruk: 1)let's not compare hashCodes in equals.
2) Can we simplify that and combine with the condition below?

TimoxaV: done

dmytro-hryhoruk (approved):

PR #141, opened 2020-08-03
Add solution

OleksiiTromsa:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +22,36 @@ public Integer getLength() {
...
+            return true;
+        }
+        if (object == null || !(object.getClass().equals(this.getClass()))) {
+            return false;
+        }
+
+        Rectangle rect = (Rectangle) object;
+        if (hashCode() != rect.hashCode()) {
+            return false;
+        }
```
KseniiaMakarova: these two methods should not explicitly rely on one another

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #144, opened 2020-08-03
Added solution

Nokafan:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+            return true;
+        }
+        if (object == null) {
+            return false;
+        }
+        if (object.getClass().equals(Rectangle.class)) {
+            Rectangle rectangle = (Rectangle) object;
+            return width == null || width.equals(rectangle.width)
+                    && length == null || length.equals(rectangle.length)
+                    && color == null || color.equals(rectangle.color);
```
KseniiaMakarova: what if width == null and rectangle.width != null?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Nokafan: Fixed. Trying to remove Main from push.

.gitignore
```diff
@@ -1,3 +1,6 @@
 .idea/*
 *.iml
 target/*
+Main.java
+MyTests.java
+TestClass.java
```
KseniiaMakarova: let's remove this
in the checklist we meant that your pushed code should not contain Main class or main() method
in other words, you should make changes in class Rectangle only

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,36 @@ public Integer getLength() {
...
+
+        if (width != null ? !width.equals(rectangle.width) : rectangle.width != null) {
+            return false;
+        }
+        if (length != null ? !length.equals(rectangle.length) : rectangle.length != null) {
+            return false;
+        }
+        if (color != null ? !color.equals(rectangle.color) : rectangle.color != null) {
+            return false;
+        }
```
KseniiaMakarova: see checklist: While comparing fields at equals, use single return statement for all fields(e.g. return check1 && check2 && check3;)

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
KseniiaMakarova: please try to improve your previous solution

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+
+        Rectangle rectangle = (Rectangle) object;
+
+        return (width == null && rectangle.width == null
+                || width != null && width.equals(rectangle.width)
+                && length == null && rectangle.length == null
+                || length != null && length.equals(rectangle.length)
+                && color == null && rectangle.color == null
+                || color != null && color.equals(rectangle.color));
```
KseniiaMakarova: The logical-AND operator ( && ) has higher precedence than the logical-OR operator ( || )
add parenthesis to this statement to ensure the flow is correct

.gitignore
```diff
@@ -1,3 +1,3 @@
 .idea/*
 *.iml
-target/*
+target/*
```
KseniiaMakarova: add black line in the end
```suggestion
target/*

```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Nokafan: It's correct I've made it first with lots oh parenthesis then slowly removed one by one redundant parenthesis. It's working exactly as it should be I've made tests for this with a null parameter for one, for two, for both and it works. But I could add some parenthesis for better understanding logic?

.gitignore
Nokafan: It should be 4 lines? 3 with text and one blank?

KseniiaMakarova (approved):

PR #145, opened 2020-08-03
added solution

Rudiachenko:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,15 +16,30 @@ public Rectangle(Integer width, Integer length, String color) {
...
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+        if (object == null || getClass() != object.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
+        return ((width == null && rectangle.width == null)
```
```suggestion
        return (width == rectangle.width
```
just a note: this can be simplified

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -16,15 +16,30 @@ public Rectangle(Integer width, Integer length, String color) {
...
-    public Integer getLength() {
-        return length;
-    }
-
-    public String getColor() {
-        return color;
+    @Override
+    public int hashCode() {
+        int result = 11;
+        int factor = 22;
```
KseniiaMakarova: it is better to use prime numbers everywhere

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #149, opened 2020-08-03
Homework 6.2 completed

Borisko130:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,39 @@ public Integer getLength() {
...
+
+        if (obj.getClass().equals(Rectangle.class)) {
+            Rectangle rect = (Rectangle) obj;
+
+            return (this.getWidth() == null && ((Rectangle) obj).getWidth() == null
+                    || width.equals(rect.getWidth()))
+                    && (this.getLength() == null && ((Rectangle) obj).getLength() == null
+                    || length.equals(rect.getLength()))
+                    && (this.getColor() == null && ((Rectangle) obj).getColor() == null
+                    || color.equals(rect.getColor()));
```
dmytro-hryhoruk: 1) Should we use getters inside class?
2) You already defined rect object, no need repeat this in every condition;
3) Try to simplify this condition

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,39 @@ public Integer getLength() {
...
+        }
+
+        if (obj == null) {
+            return false;
+        }
+
+        if (obj.getClass().equals(Rectangle.class)) {
+            Rectangle rect = (Rectangle) obj;
+
+            return (width == null && rect.width == null
```
dmytro-hryhoruk: what if only width will be null but rect.width don't?

dmytro-hryhoruk (approved):

PR #151, opened 2020-08-03
solution added

dariiabutorina:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,45 @@ public Integer getLength() {
...
+            return this.hasAllEmptyParameters()
+                    && rectangle.hasAllEmptyParameters();
+        }
+
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        if (width != null && length != null && color != null) {
```
KseniiaMakarova: if even one field is null, this method returns 0

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,45 @@ public Integer getLength() {
...
+        if (object.getClass() == Rectangle.class) {
+            Rectangle rectangle = (Rectangle) object;
+            if (!this.hasAllEmptyParameters()
+                    && !rectangle.hasAllEmptyParameters()) {
+                return width.equals(rectangle.width)
+                        && length.equals(rectangle.length)
+                        && color.equals(rectangle.color);
+            }
+            return this.hasAllEmptyParameters()
+                    && rectangle.hasAllEmptyParameters();
```
KseniiaMakarova: see checklist: `While comparing fields at equals, use single return statement for all fields(e.g. return check1 && check2 && check3;).`

KseniiaMakarova: also, I think we don't need to write a helper method

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,37 @@ public Integer getLength() {
...
+
+        if (object.getClass() == Rectangle.class) {
+            Rectangle rectangle = (Rectangle) object;
+            boolean check1 = width != null && color != null && length != null
+                    && rectangle.width != null && rectangle.color != null && rectangle.length != null;
+            boolean check2 = width == null && color == null && length == null
+                    && rectangle.width == null && rectangle.color == null && rectangle.length == null;
+            return check1 ? (width.equals(rectangle.width)
+                    && length.equals(rectangle.length)
+                    && color.equals(rectangle.color)) : check2;
```
KseniiaMakarova: please try to write this logic in a single return statement

KseniiaMakarova (changes requested):
checkstyle fails :(

KseniiaMakarova (approved):

PR #152, opened 2020-08-03
added solution

Martynice:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,26 @@ public Integer getLength() {
...
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+        if (object == null || getClass() != object.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
+        return width == null && length == null && color == null
+                ? rectangle.color == null : color.equals(rectangle.color);
```
KseniiaMakarova: why are you checking only color field here?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Martynice: changed

KseniiaMakarova (approved):

PR #153, opened 2020-08-03
homeWork6.2

khOksana99:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+                return true;
+            }
+            Rectangle rectangle = (Rectangle) object;
+            return (width == rectangle.width
+                    || (width != null && width.equals(rectangle.width)))
+                    && (length == rectangle.length
+                    || (length != null && length.equals(rectangle.length)))
+                    && (color == rectangle.color
+                    || (color != null && color.equals(rectangle.color)));
+        } else {
```
KseniiaMakarova: do we need else after return?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+            return false;
+        }
+    }
+
+    @Override
+    public int hashCode() {
+        final int variable = 5;
```
```suggestion
        final int number = 5;
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,32 @@ public Integer getLength() {
...
+        int result = 10;
```
KseniiaMakarova: let's use prime numbers everywhere

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
khOksana99: done

src/main/java/core/basesyntax/Rectangle.java
khOksana99: done

src/main/java/core/basesyntax/Rectangle.java
khOksana99: done

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object object) {
+        if (object.getClass().equals(Rectangle.class) && (object != null)) {
```
KseniiaMakarova: just noticed
will this line work properly if object == null?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
khOksana99: Yes, it will work properly, I have reakised that after the lection, so I have changed it

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object object) {
+        if (object.getClass().equals(Rectangle.class)) {
```
KseniiaMakarova, line 36: bring null check back, but put it before class check

if `object` is null, you will get NullPointerException when executing `object.getClass()`

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
khOksana99: done
We have to check if object is null, but before getClass check, right?

KseniiaMakarova (approved):

PR #156, opened 2020-08-03
homework 6-2

mikhail771:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (obj.getClass() != obj.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) obj;
+        return (width == rectangle.width || width.equals(rectangle.width))
```
dmytro-hryhoruk: what if width will be null(or other fields)?

mikhail771: done

dmytro-hryhoruk (approved):

PR #157, opened 2020-08-03
added solution

vremere-a:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,35 @@ public Integer getLength() {
...
         return color;
     }
+
+    @Override
+    public int hashCode() {
+        int result = 17;
+        final int prime = 31;
+        result = prime * result + Objects.hashCode(width);
+        result = prime * result + Objects.hashCode(length);
+        result = prime * result + Objects.hashCode(color);
```
dmytro-hryhoruk: https://mate-academy.github.io/jv-program-fulltime/01_core/checklist/hw_06.html check about using Objects

vremere-a: done

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,37 @@ public Integer getLength() {
...
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+
+        if (obj == null) {
+            return false;
+        }
+
+        if (obj instanceof Rectangle) {
```
dmytro-hryhoruk: what the difference between instanceof and getClass()?

vremere-a: instanceof use when we use extends

dmytro-hryhoruk (approved):

PR #158, opened 2020-08-03
hw6.2

OlessiaKharchenko:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -17,14 +17,45 @@ public Rectangle(Integer width, Integer length, String color) {
     }
 
     public Integer getWidth() {
+
```
KseniiaMakarova: we don't need these blank lines

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -17,14 +17,45 @@ public Rectangle(Integer width, Integer length, String color) {
...
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+        if (obj == null || obj.getClass() != this.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) obj;
+        return width == rectangle.width || (width != null
+                && width.equals(rectangle.getWidth())
```
```suggestion
        return (width == rectangle.width || (width != null
                && width.equals(rectangle.getWidth()))
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
OlessiaKharchenko: ok, done!

src/main/java/core/basesyntax/Rectangle.java
OlessiaKharchenko: done!

KseniiaMakarova (approved):

PR #159, opened 2020-08-03
solution

Vas35:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +28,31 @@ public Integer getLength() {
...
+        }
+        Rectangle rectangle = (Rectangle) obj;
+        if ((width == null && length == null && color == null)) {
+            return true;
+        }
+        if ((width == rectangle.width) && (length == rectangle.length)
+                && (color.equals(rectangle.getColor()))) {
+            return true;
+        }
+        return true;
```
dmytro-hryhoruk: should we return true if no conditions from above gave us true?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +28,31 @@ public Integer getLength() {
...
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) obj;
+        if ((width == null && length == null && color == null)) {
+            return true;
+        }
+        if ((width == rectangle.width) && (length == rectangle.length)
+                && (color.equals(rectangle.getColor()))) {
+            return true;
```
dmytro-hryhoruk: let's combine them

src/main/java/core/basesyntax/Rectangle.java
Vas35: No

src/main/java/core/basesyntax/Rectangle.java
Vas35: Ok

dmytro-hryhoruk (approved):

PR #161, opened 2020-08-04
Overrided equals & hashcode for Rectangle class

andriylagun:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+        if (!(object instanceof Rectangle) || object == null) {
```
dmytro-hryhoruk: what the difference between instanceof and getClass()?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
+        return (length == null || rectangle.length == null)
```
```suggestion
        return length == null
```
this should be enough

src/main/java/core/basesyntax/Rectangle.java
andriylagun: I see my mistake, if i will create subclass from Rectangle, object of this subclass will not equals to Rectangle object and Rectangle will be equal to child-class object. For strict type matching I need to use getClass()

dmytro-hryhoruk (commented):
1) don't call your commit "Some fixes".Let's give descriptive names to commit;
2) fix travis;

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+                && (color == null || rectangle.color == null)
+                ? color == rectangle.color : color.equals(rectangle.color);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 51;
+        result = 31 * result + (getLength() != null ? getLength().hashCode() : 0);
+        result = 31 * result + (getWidth() != null ? getWidth().hashCode() : 0);
+        result = 31 * result + (getColor() != null ? getColor().hashCode() : 0);
```
dmytro-hryhoruk, line 53: should we use getters inside class?

dmytro-hryhoruk (approved):

PR #162, opened 2020-08-04
Initial commit.

SetMyDream:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +22,34 @@ public Integer getLength() {
...
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (obj.getClass().equals(Rectangle.class)) {
+            Rectangle rectangle = (Rectangle) obj;
+            return (width == null && rectangle.width == null
```
dmytro-hryhoruk: try to simplify this condition

SetMyDream: in fact, IDE proposed me next code:
            return (Objects.equals(width, rectangle.width))
                    && (Objects.equals(length, rectangle.length))
                    && (Objects.equals(color, rectangle.color)); 
but if i`ll use it as is, it`s only checks if links are equal, right?

dmytro-hryhoruk (approved):

PR #163, opened 2020-08-04
hw

YaTut1901:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,30 @@ public Integer getLength() {
...
+        if (this == object) {
+            return true;
+        }
+        if (object == null || getClass() != object.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
+        return Objects.equals(width, rectangle.width)
+                && Objects.equals(length, rectangle.length)
+                && Objects.equals(color, rectangle.color);
```
KseniiaMakarova: please see checklist

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,30 @@ public Integer getLength() {
...
+    }
+
+    @Override
+    public int hashCode() {
+        if (width == null || length == null || color == null) {
+            return 0;
+        }
```
KseniiaMakarova: if any of your fields is null, you ignore all the others and return null
this isn't right

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +29,30 @@ public Integer getLength() {
...
+        int result = 22;
```
KseniiaMakarova: let's use prime numbers

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
YaTut1901: but it`s required to return int not null

KseniiaMakarova: sorry, I meant 
`if any of your fields is null, you ignore all the others and return 0
this isn't right`

the logic itself is not correct
you should decide whether to change `result` after checking each field, and not them all together

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+        if (this == object) {
+            return true;
+        }
+        if (object == null || getClass() != object.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
+        return width != null && width.equals(rectangle.width) || rectangle.width == width
+                && length != null && length.equals(rectangle.length) || rectangle.length == length
+                && color != null && color.equals(rectangle.color) || rectangle.color == color;
```
KseniiaMakarova: The logical-AND operator ( && ) has higher precedence than the logical-OR operator ( || )
add parenthesis to this statement to ensure the flow is correct

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,30 @@ public Integer getLength() {
...
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 17;
+        int modifier1 = width == null ? 1 : width;
+        int modifier2 = length == null ? 1 : length;
+        int modifier3 = color == null ? 1 : color.hashCode();
+        result = 31 * result * modifier1;
```
```suggestion
        result = 31 * result * width == null ? 1 : width.hashCode();
```
please refactor all three lines like this - remove `modifier` variables and use hashCode of the fields

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #164, opened 2020-08-04
redefine basic Equals and hashcode for object rectangle

fyarkiy:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,39 @@ public Integer getLength() {
...
     }
+
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+        if (object == null) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
```
KseniiaMakarova: will cast be successful if `object` is not a Rectangle?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,39 @@ public Integer getLength() {
...
+                    && ((getColor() == null && rectangle.getColor() == null)
+                    || (getColor().equals(rectangle.getColor()))));
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 23;
+        if (getWidth() != null) {
```
KseniiaMakarova: do we need to use getters from the same class?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,39 @@ public Integer getLength() {
...
+        }
+        Rectangle rectangle = (Rectangle) object;
+
+        if (object.getClass().equals(Rectangle.class)) {
+            return (((getWidth() == null && rectangle.getWidth() == null)
+                    || getWidth().equals(rectangle.getWidth()))
+                    && ((getLength() == null && rectangle.getWidth() == null)
+                    || getLength().equals(rectangle.getLength()))
+                    && ((getColor() == null && rectangle.getColor() == null)
+                    || (getColor().equals(rectangle.getColor()))));
```
KseniiaMakarova: what if width == null bit rectangle.width != null?
`getWidth().equals(rectangle.getWidth())` will this code work?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
fyarkiy: No. made changes...

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,48 @@ public Integer getLength() {
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 23;
+        if (width != null) {
+            result = 31 * result * width;
```
KseniiaMakarova: let's use width.hashCode() here

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,48 @@ public Integer getLength() {
...
+        }
+        if (length != null) {
+            result = 31 * result * length;
```
KseniiaMakarova: same

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,48 @@ public Integer getLength() {
...
+                    || (getColor() != null && rectangle.getColor() == null)) {
+                return false;
+            }
+
+            return (((getWidth() == null && rectangle.getWidth() == null)
+                    || getWidth().equals(rectangle.getWidth()))
+                    && ((getLength() == null && rectangle.getWidth() == null)
+                    || getLength().equals(rectangle.getLength()))
+                    && ((getColor() == null && rectangle.getColor() == null)
+                    || (getColor().equals(rectangle.getColor()))));
```
KseniiaMakarova: move all this logic to a single return statement please
see checklist: While comparing fields at equals, use single return statement for all fields(e.g. return check1 && check2 && check3;)

also, don't use getters - use fields instead. you have direct access to the fields from the class where they are declared

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #167, opened 2020-08-04
task is done

yutsevich:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,29 @@ public Integer getLength() {
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return ((rectangle.width != null && width.equals(rectangle.width))
```
```suggestion
        return ((width != null && width.equals(rectangle.width))
```
same for all the fields
reason: if width is null, this line in your solution would throw NullPointerException

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #168, opened 2020-08-04
solution added

Vosker:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
...
+            return true;
+        }
+        if (obj == null || !obj.getClass().equals(Rectangle.class)) {
+            return false;
+        }
+        Rectangle rect = (Rectangle) obj;
+
+        return width == null
+                ? width == rect.width : width.equals(width)
+                && (length == null || rect.length == null)
```
dmytro-hryhoruk: can we do here the same check as with width?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public int hashCode() {
+        int result = 61;
+        result = 31 * result + (getLength() == null ? 0 : getLength().hashCode());
+        result = 31 * result + (getWidth() == null ? 0 : getWidth().hashCode());
+        result = 31 * result + (getColor() == null ? 0 : getColor().hashCode());
```
dmytro-hryhoruk: should we use getters inside class?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -27,4 +27,31 @@ public Integer getLength() {
...
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || !obj.getClass().equals(Rectangle.class)) {
+            return false;
+        }
+        Rectangle rect = (Rectangle) obj;
+
+        return (length != null && rect.length != null)
```
dmytro-hryhoruk: do we need to check both length on null?

dmytro-hryhoruk (commented):
fix travis

dmytro-hryhoruk (approved):

PR #182, opened 2020-12-06
equals and hashcode by Oleh Kravets

Ketch92:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+            return false;
+        }
+        if ((object.getClass().equals(Rectangle.class))) {
+            Rectangle rectangle = (Rectangle) object;
+            return (rectangle.length == null && this.length == null
+                    || this.length.compareTo(rectangle.length) == 0)
+                    && (rectangle.color == null && this.color == null
+                    || this.color.equals(rectangle.color))
+                    && (rectangle.width == null && this.width == null
+                    || this.width.compareTo(rectangle.width) == 0);
```
```suggestion
            return (rectangle.length == null && this.length == null
                    || this.length.compareTo(rectangle.length) == 0)
                    && (rectangle.color == null && this.color == null
                    || this.color.equals(rectangle.color))
                    && (rectangle.width == null && this.width == null
                    || this.width.compareTo(rectangle.width) == 0);
```
There is no need use this keyword if you are calling the field of this class . We should use this keyword in constructors and setters when the method parameter and the field name are the same. But here it's redundant.

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+        if (object == this) {
+            return true;
+        }
+        if (object == null) {
+            return false;
+        }
+        if ((object.getClass().equals(Rectangle.class))) {
+            Rectangle rectangle = (Rectangle) object;
+            return (rectangle.length == null && length == null
+                    || length.compareTo(rectangle.length) == 0)
```
dmytro-hryhoruk: 1)You have a potential NPE risk here when length == null, but rectangle.length is not;
2) use equals() instead of compareTo

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Ketch92: I have seen your post in Slack about this and removed them, but maybe I forgot to push new version

Ketch92: I did remove all thises and pushed (you can check this commit above in the list), dunno how it was in the code.

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+        if (object == this) {
+            return true;
+        }
+        if (object == null) {
+            return false;
+        }
+        if ((object.getClass().equals(Rectangle.class))) {
+            Rectangle rectangle = (Rectangle) object;
+            return (rectangle.length == null && length == null
+                    || length.equals(rectangle.length))
```
```suggestion
            return (rectangle.length == null && length == null
                    || length.equals(rectangle.length))
```
rectangle.length != null, length == null —> you will get NPE by this condition
In this case this check (rectangle.length == null) will give us false and we will gonna move to the right part length.equals(rectangle.length)) where NPE will happen;

dmytro-hryhoruk (changes requested):

dmytro-hryhoruk (approved):

PR #184, opened 2020-12-07
equals-hashcode task done

less08:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -1,5 +1,7 @@
 package core.basesyntax;
 
+import java.util.Objects;
```
boroda4436: please see the common mistakes file

boroda4436 (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,44 @@ public Integer getLength() {
...
+            Rectangle current = (Rectangle) obj;
+            return compareField(this.width, current.width)
+                    && compareField(this.length, current.length)
+                    && compareField(this.color, current.color);
+        }
+        return false;
+    }
+
+    private boolean compareField(Integer first, Integer second) {
+        if (first == null && second != null || first != null && second == null) {
```
boroda4436: can we simplify this method?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,44 @@ public Integer getLength() {
...
+            return false;
+        }
+        return first == null || first.equals(second);
+    }
+
+    private boolean compareField(String first, String second) {
```
boroda4436: code duplication with the previous method. I guess we can avoid writing `compareField` methods. We can write all logic in `equals` methods

boroda4436 (changes requested):

KseniiaMakarova (approved):

PR #186, opened 2020-12-07
methods equals() and hashCode() implemented into Rectangle class

semyonich:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,30 @@ public Integer getLength() {
...
+            return true;
+        }
+        if (rectangle == null) {
+            return false;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle current = (Rectangle) rectangle;
+            return (width == current.getWidth() || width.equals(current.getWidth()))
+                    && (length == current.getLength() || length.equals(current.getLength()))
+                    && (color == current.getColor() || color.equals(current.getColor()));
```
KseniiaMakarova: what if `width` is null and `current.width` is not null?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
semyonich: done

KseniiaMakarova (approved):

PR #187, opened 2020-12-07
Override methods hashCode() and equals()

YuraMitriakhin:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +24,27 @@ public Integer getLength() {
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return Objects.equals(width, rectangle.width)
+                && Objects.equals(length, rectangle.length)
+                && Objects.equals(color, rectangle.color);
```
KseniiaMakarova: please see the checklist

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,27 @@ public Integer getLength() {
...
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return ((width == null && rectangle.width == null) || width.equals(rectangle.width))
+                && ((length == null && rectangle.width == null) || length.equals(rectangle.length))
+                && ((color == null && color == null) || color.equals(rectangle.color));
+    }
+
+    @Override
+    public int hashCode() {
+        int results = 17;
```
```suggestion
        int result = 17;
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,27 @@ public Integer getLength() {
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) o;
+        return ((width == null && rectangle.width == null) || width.equals(rectangle.width))
+                && ((length == null && rectangle.width == null) || length.equals(rectangle.length))
+                && ((color == null && color == null) || color.equals(rectangle.color));
```
KseniiaMakarova: what if `width` is null and `rectangle.width` is not null?

also, if both of them are null, the following code will return true `width == rectangle.width`

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #188, opened 2020-12-07
overrided equals and hashcode methods

Alex-Korzun:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,30 @@ public Integer getLength() {
...
+    @Override
+    public boolean equals(Object rectangle) {
+        if (this == rectangle) {
+            return true;
+        }
+        if (rectangle == null || rectangle.getClass() != this.getClass()) {
+            return false;
+        }
+        Rectangle castedRectangle = (Rectangle) rectangle;
+        return (width == (castedRectangle.width)
```
```suggestion
        return (width == castedRectangle.width
```
remove for other fields too

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,30 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object rectangle) {
+        if (this == rectangle) {
+            return true;
+        }
+        if (rectangle == null || rectangle.getClass() != this.getClass()) {
```
```suggestion
        if (rectangle == null || rectangle.getClass() != getClass()) {
```

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Alex-Korzun: removed braces

src/main/java/core/basesyntax/Rectangle.java
Alex-Korzun: removed `this`

dmytro-hryhoruk (approved):

PR #189, opened 2020-12-07
HW - 6 (equals & hashCode)

Baldinini:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,32 @@ public Integer getLength() {
...
+            return true;
+        }
+        if (rectangle == null) {
+            return false;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle firstRectangle = (Rectangle) rectangle;
+            return ((width == null) || width.equals(firstRectangle.width))
+                    && (length == null || length.equals(firstRectangle.length))
+                    && (color == null || color.equals(firstRectangle.color));
```
dmytro-hryhoruk: what if width of current object  == null, but firstRectangle.width has some other value? Should the result be true?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,32 @@ public Integer getLength() {
...
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 11;
+
+        result = result + (width == null ? 0 : width);
+        result = result + (length == null ? 0 : length);
+        result = result + (color == null ? 0 : color.hashCode());
+
```
```suggestion
        result = result + (width == null ? 0 : width);
        result = result + (length == null ? 0 : length);
        result = result + (color == null ? 0 : color.hashCode());
```

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+        if (rectangle == this) {
+            return true;
+        }
+        if (rectangle == null) {
+            return false;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle firstRectangle = (Rectangle) rectangle;
+            return ((width == null && firstRectangle.width == null)
+                    || width.equals(firstRectangle.width))
```
dmytro-hryhoruk: NPE here when width == null and firstRectangle.width is not

dmytro-hryhoruk (changes requested):

dmytro-hryhoruk (approved):

PR #191, opened 2020-12-07
Hm equals hash code

tonitrus55:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,32 @@ public Integer getLength() {
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int hashCode = 17;
+        hashCode *= 31 + (this.length == null ? 0 : this.length);
+        hashCode *= 31 + (this.color == null ? 0 : this.color.length());
+        hashCode *= 31 + (this.width == null ? 0 : this.width);
```
KseniiaMakarova: let's not use `this` keyword - it is redundant

also, please use hashcode of the field instead of the string length

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,32 @@ public Integer getLength() {
...
+            return true;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle compareRectangle = (Rectangle) rectangle;
+            return (this.width == compareRectangle.getWidth()
+                    || this.width.equals(compareRectangle.getWidth()))
+                    && (this.length == compareRectangle.getLength()
+                    || this.length.equals(compareRectangle.getLength()))
+                    && (this.color == compareRectangle.getColor()
+                    || this.color.equals(compareRectangle.getColor()));
```
KseniiaMakarova: let's not use `this`

and what if `width` is null but `compareRectangle.getWidth()` is not null?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
tonitrus55: Я исправлю, но у меня вопрос по this:
Почему нельзя? 
Я, например, считаю что this. увеличивает читаемость кода.

KseniiaMakarova: https://stackoverflow.com/questions/2411270/when-should-i-use-this-in-a-class
in this conversation you can read about this issue

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,32 @@ public Integer getLength() {
...
+        if (rectangle == null) {
+            return false;
+        }
+        if (rectangle == this) {
+            return true;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle compareRectangle = (Rectangle) rectangle;
+            return ((width == null && compareRectangle.getWidth() == null)
+                    || width.equals(compareRectangle.getWidth()))
```
KseniiaMakarova: again, what if `width` is null but `compareRectangle.getWidth()` is not null?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,32 @@ public Integer getLength() {
...
+    public boolean equals(Rectangle rectangle) {
+        if (rectangle == null) {
+            return false;
+        }
+        if (rectangle == this) {
+            return true;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle compareRectangle = (Rectangle) rectangle;
+            return ((width == null && compareRectangle.getWidth() == null)
```
```suggestion
            return (width == compareRectangle.getWidth()
```
this can be replaced

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,32 @@ public Integer getLength() {
...
+            return true;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle compareRectangle = (Rectangle) rectangle;
+            return ((width == compareRectangle.width)
+                    || (width != null && width.equals(compareRectangle.width)))
+                    && (length == compareRectangle.length)
+                    || (length != null && length.equals(compareRectangle.getLength()))
+                    && (color == compareRectangle.getColor()
+                    || (color != null && color.equals(compareRectangle.getColor())));
```
```suggestion
            return (width == compareRectangle.width
                    || (width != null && width.equals(compareRectangle.width)))
                    && (length == compareRectangle.length
                    || (length != null && length.equals(compareRectangle.length)))
                    && (color == compareRectangle.color
                    || (color != null && color.equals(compareRectangle.color)));
```

KseniiaMakarova (changes requested):
fix parenthesis so that the parts of logical construction were executed in the right order

KseniiaMakarova (approved):

PR #196, opened 2020-12-07
added solution equals-hashcode

IgorBochkor:
added solution equals-hashcode

### DIFF
--- a/src/main/java/core/basesyntax/Rectangle.java
+++ b/src/main/java/core/basesyntax/Rectangle.java
@@ -35,12 +35,12 @@ public class Rectangle {
 
         if (rectangle.getClass().equals(Rectangle.class)) {
             Rectangle current = (Rectangle) rectangle;
-            return (this.width == current.width || this.width != null
-                    && this.width.equals(current.width))
-                    && (this.color == current.color || this.color != null
-                    && this.color.equals(current.color))
-                    && (this.length == current.length || this.length != null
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,36 @@ public Integer getLength() {
...
+    @Override
+    public boolean equals(Object rectangle) {
+        if (rectangle == null) {
+            return false;
+        }
+
+        if (this == rectangle) {
+            return true;
+        }
+
```
dmytro-hryhoruk, line 35:
```suggestion
        if (rectangle == null) {
            return false;
        }
        if (this == rectangle) {
            return true;
        }
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,36 @@ public Integer getLength() {
...
+    @Override
+    public int hashCode() {
+        int result = 19;
+        result = 37 * (width == null ? 0 : width.hashCode());
+        result = 37 * (color == null ? 0 : color.hashCode());
+        result = 37 * (length == null ? 0 : length.hashCode());
+        return result;
+    }
 }
+
```
dmytro-hryhoruk, line 57:
```suggestion
```

dmytro-hryhoruk (approved):

PR #198, opened 2020-12-07
add solution for override equals and hashCode methods;

mzrkvolodymyr:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+            return false;
+        }
+        if (rectangle.getClass().equals(this.getClass())) {
+            Rectangle current = (Rectangle) rectangle;
+            return (this.width == current.width || this.width != null
+                    && this.width.equals(current.width))
+                    && (this.length == current.length || this.length != null
+                    && this.length.equals((current.length)))
+                    && (this.color == current.color || this.color != null
+                    && this.color.equals(current.color));
```
KseniiaMakarova: we don't have to use `this` keyword here - it is redundant

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
mzrkvolodymyr: thanks
done

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+        if (rectangle == null) {
+            return false;
+        }
+        if (rectangle.getClass().equals(this.getClass())) {
+            Rectangle current = (Rectangle) rectangle;
+            return (width == current.width || width != null
+                    && width.equals(current.width))
+                    && (length == current.length || length != null
+                    && length.equals((current.length)))
+                    && (color == current.color || this.color != null
```
KseniiaMakarova, line 40:
```suggestion
                    && (color == current.color || color != null
```

KseniiaMakarova (approved):

PR #199, opened 2020-12-07
added solution

mfedenko09:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,32 @@ public Integer getLength() {
...
+    public boolean equals(Object rectangle) {
+        if (rectangle == this) {
+            return true;
+        }
+        if (rectangle == null) {
+            return false;
+        }
+        if (this.color == null && this.width == null && this.length == null) {
+            return true;
+        }
```
KseniiaMakarova: are you sure that this check is needed?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,32 @@ public Integer getLength() {
...
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle current = (Rectangle) rectangle;
+            return this.color.equals(current.color) && this.length.equals(current.length)
+                    && this.width.equals(current.width);
```
KseniiaMakarova: let's not use `this` keyword - it is redundant
and let's refactor this check

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
mfedenko09: yes, I am, because when 3 fields equals null, I should return true and equal hashCode()
image
image
image

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,30 @@ public Integer getLength() {
...
+            return true;
+        }
+        if (rectangle == null) {
+            return false;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle current = (Rectangle) rectangle;
+            return (color == current.color && length == current.length
+                    && width == current.width) || (color.equals(current.color)
+                    && length.equals(current.length) && width.equals(current.width));
```
KseniiaMakarova: let’s follow the following logical flow for each field:

return true,

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,32 @@ public Integer getLength() {
...
+        if (rectangle == null) {
+            return false;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle current = (Rectangle) rectangle;
+            return (color == null && current.color == null && length == null
+                    && current.length == null && width == null && current.width == null)
+                    || (color != null && length != null && width != null
+                    && color.equals(current.color) && length.equals(current.length)
+                    && width.equals(current.width));
```
```suggestion
            return ((color == null && current.color == null) 
            || (color != null && color.equals(current.color))
            && (same for width)
            && (same for length)
            
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+            return false;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle current = (Rectangle) rectangle;
+            return ((color == null && current.color == null)
+                    || (color != null && color.equals(current.color))
+                    && ((length == null && current.length == null)
+                    || (length != null && length.equals(current.length))
+                    && ((width == null && current.width == null)
+                    || (width != null && width.equals(current.width)))));
```
```suggestion
            return ((color == null && current.color == null)
                    || (color != null && color.equals(current.color)))
                    && ((length == null && current.length == null)
                    || (length != null && length.equals(current.length)))
                    && ((width == null && current.width == null)
                    || (width != null && width.equals(current.width)));
```

KseniiaMakarova (changes requested):
please fix parenthesis - now the order of statement execution is not correct

dmytro-hryhoruk (approved):

KseniiaMakarova (approved):

PR #202, opened 2020-12-07
added equals and hashCode to class Rectangle

Alekearl:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +24,32 @@ public Integer getLength() {
...
+    @Override
+    public boolean equals(Object object) {
+        if (object == null) {
+            return false;
+        }
+
+        if (this == object) {
+            return true;
+        }
+
```
```suggestion
        if (object == null) {
            return false;
        }
        if (this == object) {
            return true;
        }
```

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +24,32 @@ public Integer getLength() {
...
+        if (object.getClass().equals(Rectangle.class)) {
+            Rectangle figure = (Rectangle) object;
+            return Objects.equals(this.width, figure.width)
+                    && Objects.equals(this.length, figure.length)
+                    && Objects.equals(this.color, figure.color);
```
dmytro-hryhoruk: https://mate-academy.github.io/jv-program-common-mistakes/java-core/equals-and-hashcode/equals-and-hashcode

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Alekearl: Removed methods of Objects.class.

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,30 @@ public Integer getLength() {
...
+    public boolean equals(Object object) {
+        if (object == null) {
+            return false;
+        }
+        if (this == object) {
+            return true;
+        }
+        if (object.getClass().equals(Rectangle.class)) {
+            Rectangle figure = (Rectangle) object;
+            return (width == figure.width || width.equals(figure.width))
```
dmytro-hryhoruk: what if width = null, but figure.width is not, then we will have NPE in the right part

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Alekearl: Solved, i think.  =) 
In this last commit IDE wants to replace it with Objects.equals()...

dmytro-hryhoruk (approved):

PR #204, opened 2020-12-07
Initial

Max-Pochepets:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,27 @@ public Integer getLength() {
...
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        Rectangle current = (Rectangle) obj;
+        return (this.width == null || this.width.equals(current.width))
+                && (this.length == null || this.length.equals(current.length))
+                && (this.color == null || this.color.equals(current.color));
```
KseniiaMakarova: are you sure that this check is ok?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Max-Pochepets: we can't invoke equals() on null object, so we need to check for it

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,29 @@ public Integer getLength() {
...
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        Rectangle current = (Rectangle) obj;
+        return (width == null && current.width == null || width.equals(current.width))
+                && (length == null && current.length == null
+                    || length.equals(current.length))
+                && (color == null && current.color == null
+                    || color.equals(current.color));
```
KseniiaMakarova: what if `width` is null and `current.width` is not null? what will we get on line 35?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,30 @@ public Integer getLength() {
...
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        Rectangle current = (Rectangle) obj;
+        return (width == null && current.width == null
+                    || (width != null && width.equals(current.width)))
+                && (length == null && current.length == null
+                    || (length != null && length.equals(current.length)))
+                && (color == null && current.color == null
+                    || (color != null && color.equals(current.color)));
```
```suggestion
        return (width == current.width
                    || (width != null && width.equals(current.width)))
                && (length == current.length
                    || (length != null && length.equals(current.length)))
                && (color == current.color
                    || (color != null && color.equals(current.color)));
```
this can be simplified a bit

KseniiaMakarova (approved):

src/main/java/core/basesyntax/Rectangle.java
Max-Pochepets: :+1:

PR #205, opened 2020-12-07
made solution

Sashk52:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,40 @@ public Integer getLength() {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+
+        Rectangle rectangle = (Rectangle)obj;
```
KseniiaMakarova: what will happen on this line if `obj` is of type String?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,40 @@ public Integer getLength() {
...
+            }
+        }
+        if (color == null && rectangle.color == null && width == null && rectangle.width == null
+                && length == null && rectangle.length == null) {
+            return true;
+        }
+
+        if (obj.getClass().equals(Rectangle.class)) {
+            return rectangle.getWidth().equals(this.getWidth()) && rectangle.getLength()
+                    .equals(this.getLength()) && rectangle.getColor().equals(this.getColor());
```
KseniiaMakarova: let's move null checks here as well (try to write everything in a single statement after you cast the `obj`

also, please do not use `this` keyword when addressing the fields if the current object - it is redundant

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Sashk52: ClassCastException?

Sashk52: Should i use try-catch for casting?

src/main/java/core/basesyntax/Rectangle.java
Sashk52: done

src/main/java/core/basesyntax/Rectangle.java
Sashk52: done

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,45 @@ public Integer getLength() {
...
+                        || rectangle.getWidth().equals(getWidth()))
+                        && ((width == null && rectangle.width == null)
+                        || rectangle.getLength().equals(getLength()))
+                        && ((length == null && rectangle.length == null)
+                        || rectangle.getColor().equals(getColor()));
+            }
+
+        } catch (ClassCastException e) {
+            throw new RuntimeException("obj can't be casted to Rectangle", e);
+        }
```
KseniiaMakarova: let's not use try-catch

you already have type check on line 42
the problem is that you cast before this check

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,45 @@ public Integer getLength() {
...
+                    return false;
+                }
+            }
+            if (obj.getClass().equals(Rectangle.class)) {
+                return ((color == null && rectangle.color == null)
+                        || rectangle.getWidth().equals(getWidth()))
+                        && ((width == null && rectangle.width == null)
+                        || rectangle.getLength().equals(getLength()))
+                        && ((length == null && rectangle.length == null)
+                        || rectangle.getColor().equals(getColor()));
```
KseniiaMakarova: let’s follow the following logical flow for each field:

return true,

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Sashk52: but if cast after checking, i have no access to rectangle fields of Object, because it's not an Rectangle yet, and it has no field width, for example...that's why i can't check.

src/main/java/core/basesyntax/Rectangle.java
Sashk52: done (added additional checking for each fileds if it is not null)

src/main/java/core/basesyntax/Rectangle.java
Sashk52: I guess i have got: i must cast exactly after cheking object type. Done.

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,37 @@ public Integer getLength() {
...
+        }
+        if (obj == null) {
+            return false;
+        }
+
+        if (obj.getClass().equals(Rectangle.class)) {
+            Rectangle rectangle = (Rectangle)obj;
+            return ((color == null && rectangle.color == null)
+                        || (color != null && rectangle.color != null)
+                        && rectangle.getWidth().equals(getWidth()))
```
KseniiaMakarova: now I agree with the logic - good job!

but this is a single logic block where you compare the same field of two rectangles
you take `color` everywhere but suddenly `width` in the end

same for other fields

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Sashk52: done

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,37 @@ public Integer getLength() {
...
+            Rectangle rectangle = (Rectangle)obj;
+            return ((color == null && rectangle.color == null)
+                        || (color != null && rectangle.color != null)
+                        && rectangle.getColor().equals(getColor()))
+                        && ((width == null && rectangle.width == null)
+                        || (width != null && rectangle.width != null)
+                        && rectangle.getWidth().equals(getWidth()))
+                        && ((length == null && rectangle.length == null)
+                        || (length != null && rectangle.length != null)
+                        && rectangle.getLength().equals(getLength()));
```
KseniiaMakarova: good job

here is how we can simplify this a bit
note that we don't have to use getters because we are located inside the class

```suggestion
            return (color == rectangle.color              // both null check
                        || (color != null && color.equals(rectangle.color)))
                        && (length == rectangle.length
                        || (length != null && length.equals(rectangle.length)))
                        && (width == rectangle.width 
                        || (width != null && width.equals(rectangle.width)));
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
Sashk52: done

KseniiaMakarova (approved):

PR #206, opened 2020-12-08
done

AleksandrPok:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+
+    @Override
+    public boolean equals(Object rectangle) {
+        if (rectangle == this) {
+            return true;
+        }
+        if (rectangle == null) {
+            return false;
+        }
+        Rectangle castedRectangle = (Rectangle) rectangle;
```
KseniiaMakarova: what happens on this line if rectangle is of type String?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+        if (!castedRectangle.getClass().toString().equals(Rectangle.class.toString())) {
```
KseniiaMakarova: there is no need to call class.toString()

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+            return false;
+        }
+        return ((this.color == null && castedRectangle.color == null)
+                || this.color.equals(castedRectangle.color))
+                && ((this.width == null && castedRectangle.width == null)
+                || this.width.equals(castedRectangle.width))
+                && ((this.length == null && castedRectangle.length == null)
+                || this.length.equals(castedRectangle.length));
```
KseniiaMakarova: 1) let's not use `this`

2) if both `color` and `castedRectangle.color` is null, the following statement will be `true` - `color == castedRectangle.color`

3) what if `color` is null and `castedRectangle.color` is not null?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,30 @@ public Integer getLength() {
...
+        if (rectangle == null) {
+            return false;
+        }
+        Rectangle castedRectangle = (Rectangle) rectangle;
+        if (!castedRectangle.getClass().equals(Rectangle.class)) {
+            return false;
+        }
+        return (color == null || color.equals(castedRectangle.color))
+                && (width == null || width.equals(castedRectangle.width))
+                && (length == null || length.equals(castedRectangle.length));
```
KseniiaMakarova: let's follow the following logical flow for each field:

return true, 

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,30 @@ public Integer getLength() {
...
+
+    @Override
+    public boolean equals(Object rectangle) {
+        if (rectangle == this) {
+            return true;
+        }
+        if (rectangle == null) {
+            return false;
+        }
+        Rectangle castedRectangle = (Rectangle) rectangle;
```
KseniiaMakarova: let's protect ourselves from ClassCastException

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+        if (!rectangle.getClass().equals(Rectangle.class)) {
+            return false;
+        }
+        Rectangle castedRectangle = (Rectangle) rectangle;
+        return ((color == null && castedRectangle.color == null)
+                || (color != null && color.equals(castedRectangle.color)))
+                && ((width == null && castedRectangle.width == null)
+                || (width != null && width.equals(castedRectangle.width)))
+                && ((length == null && castedRectangle.length == null)
+                || (length != null && length.equals(castedRectangle.length)));
```
KseniiaMakarova: good job!
this may be simplified a bit

```suggestion
        return (color == castedRectangle.color
                || (color != null && color.equals(castedRectangle.color)))
                && (width == castedRectangle.width
                || (width != null && width.equals(castedRectangle.width)))
                && (length == castedRectangle.length
                || (length != null && length.equals(castedRectangle.length)));
```

KseniiaMakarova (approved):

PR #207, opened 2020-12-08
My cranch

AntonPanchishko:


KseniiaMakarova (changes requested):
fix travis please

don't push your code until you run `mvn clean package` locally

PR #208, opened 2020-12-08
overridden equals/ hashcode

AndriiBardin:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,33 @@ public Integer getLength() {
...
+        Rectangle current = (Rectangle) rectangle;
+        if (this == rectangle) {
+            return true;
+        }
+        if (rectangle == null) {
+            return false;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            return (width == null && current.width == null
+                    || width.equals(current.width))
```
dmytro-hryhoruk: what if width == null, current.width != null then we will checking the right part of ||(OR) and have NullPointerException

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/Rectangle.java
AndriiBardin: issue resolved

dmytro-hryhoruk (approved):

PR #210, opened 2020-12-08
1) Commit

KostyantynProskura:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,34 @@ public Integer getLength() {
...
+
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+        if (object == null) {
+            return false;
+        }
+        Rectangle rectangle = (Rectangle) object;
```
KseniiaMakarova: you are casing before the type check
 what if `object` is of type String?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,34 @@ public Integer getLength() {
...
+        if ((this.color == null && rectangle.color == null)
+                && (this.length == null && rectangle.length == null)
+                && (this.width == null && rectangle.width == null)) {
+            return true;
+        } else if (object.getClass().equals(Rectangle.class)) {
+            return (this.color != null && this.color.equals(rectangle.color))
+                    && (this.length != null && this.length.equals(rectangle.length))
+                    && (this.width != null && this.width.equals(rectangle.width));
```
KseniiaMakarova: let's not use `this` keyword - here it is redundant

also, let's remove lines 35-38 and follow the following logical flow for each field on lines 39-42:

return true,

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #214, opened 2020-12-08
Implement methods equals() and hashCode()

Resci:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -11,6 +11,33 @@ public Rectangle(Integer width, Integer length, String color) {
...
+    public boolean equals(Rectangle rectangle) {
+        if (this == rectangle) {
+            return true;
+        }
+        if (rectangle == null) {
+            return false;
+        }
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            return ((width == rectangle.width)
+                    || (width != null && width.equals(rectangle.width)))
```
```suggestion
            return (width == rectangle.width
                    || width != null && width.equals(rectangle.width))
```
remove redundant braces;

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -11,6 +11,33 @@ public Rectangle(Integer width, Integer length, String color) {
...
+                    || (length != null && length.equals(rectangle.length)))
+                    && ((color == rectangle.color)
+                    || (color != null && color.equals(rectangle.color)));
+        }
+        return false;
+    }
+
+    public int hashCode() {
+        if (getWidth() == null
+                || getLength() == null
```
dmytro-hryhoruk: why getters?

dmytro-hryhoruk: hashcode should be zero if length == null but width !=null;?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -11,6 +11,33 @@ public Rectangle(Integer width, Integer length, String color) {
...
+                || color == null) {
+            return 0;
+        }
+        return ((width + 17) * 23 + length) * 31 + color.hashCode();
```
dmytro-hryhoruk: let's use hashCode() method for each field

dmytro-hryhoruk (changes requested):

dmytro-hryhoruk (approved):

PR #215, opened 2020-12-08
Yegor sereda

entertainment-veks:


src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,72 @@ public Integer getLength() {
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 17;
+        result = 31 * result + (this.length == null ? 0 : this.length);
+        result = 31 * result + (this.width == null ? 0 : this.width);
+        result = 31 * result + (this.color == null ? 0 : this.color.hashCode());
```
KseniiaMakarova: let's not use `this` keyword - it is redundant here

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,72 @@ public Integer getLength() {
     public String getColor() {
         return color;
     }
+
+    @Override
+    public boolean equals(Object rectangle) {
+
+        Rectangle current = (Rectangle) rectangle;
```
KseniiaMakarova: what will be the result of this line if `rectangle` is of type String?

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,72 @@ public Integer getLength() {
...
+        if (rectangle == this) {
+            return true;
+        }
+
+        if (rectangle == null) {
+            return false;
+        }
+
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            return colorCheck(current) && lengthCheck(current) && widthCheck(current);
```
KseniiaMakarova: let's remove all those private methods and create a single statement using && and ||

follow the following logical flow for each field:

return true,

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,36 @@ public Integer getLength() {
...
+        }
+
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle current = (Rectangle) rectangle;
+            return ((current.color == null && this.color == null)
+                    || current.color.equals(this.color))
+                    && ((current.length == null && this.length == null)
+                    || current.length.equals(this.length))
+                    && ((current.width == null && this.width == null)
+                    || current.width.equals(this.width));
```
KseniiaMakarova: let's not use `this` keyword

src/main/java/core/basesyntax/Rectangle.java
```diff
@@ -22,4 +22,36 @@ public Integer getLength() {
...
+        }
+
+        if (rectangle == null) {
+            return false;
+        }
+
+        if (rectangle.getClass().equals(Rectangle.class)) {
+            Rectangle current = (Rectangle) rectangle;
+            return ((current.color == null && this.color == null)
+                    || current.color.equals(this.color))
```
KseniiaMakarova: what if `current.color` is null and `color` is not null?

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #218, opened 2020-12-14
added test cases for Equals and hashcode

dmytro-hryhoruk:


KseniiaMakarova (commented):
maybe we should also add tests for 
`Rectangle first = new Rectangle(300, 400, "red")`;
and 
`Rectangle second = new Rectangle (new Integer(300), new Integer(400), new String("red");`

you got the point - to check if `equals()` is used, not only comparison by `==`

KseniiaMakarova (approved):

